PR #177, opened 2022-11-19
some gas optimizations👷

codeislight1:


ewilz:
you need to run `yarn test:gas` to update the gas snapshots. Unfortunately these changes increase gas.

PR #178, opened 2022-11-19
small binary gas optimization for Dispatcher.sol

DennisDv24:
In the Dispatcher.sol you use the condition isNotNFTType to reduce the conditions space by 2, then you do it again with is0to7, leaving it to 8 conditions checks until the correct command is found in the worst case. I applied the same logic with a third variable "is0to3" to reduce the number of conditions by 2 again. All commands such that `commandType & 0x04 == 0` will cost a little more gas because of the extra condition, while `commandType & 0x04 != 0` will have a bigger gas upgrade than the downgrade.

### DIFF
--- a/contracts/base/Dispatcher.sol
+++ b/contracts/base/Dispatcher.sol
@@ -31,125 +31,143 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks 
     function dispatch(bytes1 commandType, bytes memory inputs) internal returns (bool success, bytes memory output) {
         bool isNotNFTType = (commandType & Commands.NFT_TYPE_MASK) == 0;
         bool is0To7 = (commandType & Commands.SUB_IF_BRANCH_MASK) == 0;
+		bool is0to3 = (commandType & 0x04) == 0;
         uint256 command = uint8(commandType & Commands.COMMAND_TYPE_MASK);
 
         success = true;
 
         if (isNotNFTType) {
             if (is0To7) {
-                if (command == Commands.V3_SWAP_EXACT_IN) {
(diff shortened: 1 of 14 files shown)


contracts/base/Dispatcher.sol
```diff
@@ -31,125 +31,143 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks
...
+						v3SwapExactInput(recipient.map(), amountIn, amountOutMin, path, payer);
+					} else if (command == Commands.V3_SWAP_EXACT_OUT) {
+						(address recipient, uint256 amountOut, uint256 amountInMax, bytes memory path, bool payerIsUser) =
+							abi.decode(inputs, (address, uint256, uint256, bytes, bool));
+						address payer = payerIsUser ? msg.sender : address(this);
+						v3SwapExactOutput(recipient.map(), amountOut, amountInMax, path, payer);
+					} else if (command == Commands.PERMIT2_TRANSFER_FROM) {
+						(address token, address recipient, uint160 amount) = abi.decode(inputs, (address, address, uint160));
+						permit2TransferFrom(token, msg.sender, recipient, amount);
+					} else { // PERMIT2_PERMIT_BATCH
```
DennisDv24, line 55: I guess in this cases i should leave the `else if` and add a final else with a `revert InvalidCommandType(command)`

ewilz:
It looks like gas overall has gone up. also plz run `yarn prettier:fix` when submitting!

DennisDv24:
> It looks like gas overall has gone up

Yeah i just checked with a script and overall its a negligible upgrade, so nevermind.

PR #182, opened 2022-11-21
feat: add lite profile

marktoda:
For quick testing, add FOUNDRY_PROFILE=lite to get ~8s compilation times instead of ~45s


ewilz:
so if we run this in universal-router-sdk, we'll need `FOUNDRY_PROFILE=lite` in our command?

ewilz (approved):

PR #183, opened 2022-11-21
Alternative nesting logic

hensha256:
Alternative to #181 - adding clear placeholders for contributors to replace with new commands. Keeps all `if` nests at 8 depth.
We can then add instructions for contributors to replace a placeholder command with their command if they want to.


ewilz:
yahh I think since this the router is supposed to be ongoing I'm into this (after going back and forth). I'm not totally sold on nesting, however, this is a good compromise. I'm sure we'll iterate on this in the future to come

script/DeployUniversalRouter.s.sol
```diff
@@ -9,7 +9,6 @@ import {UnsupportedProtocol} from 'contracts/deploy/UnsupportedProtocol.sol';
 import {UniversalRouter} from 'contracts/UniversalRouter.sol';
 import {Permit2} from 'permit2/src/Permit2.sol';
 
-/// @dev The SALT used to deploy Universal Router on Mainnet is 0x000000000000000000000000000000000000000068f501dc113220003862815d
 bytes32 constant SALT = bytes32(uint256(0x00000000000000000000000000000000000000005eb67581652632000a6cbedf));
```
snreynolds, line 12: Is this the salt you deployed with?

snreynolds (approved):

script/DeployUniversalRouter.s.sol
hensha256: yeah

Issue #186, opened 2022-11-22
NPM package cannot find Permit2 library


PowVT:
I installed the UR repo with this command `yarn add @uniswap/universal-router`. 

Then added the UniversalRouter.sol contract as an import into a contract to compile and generate its artifacts. 

The first error I see is `Error HH411: The library solmate, imported from @uniswap/universal-router/contracts/base/Dispatcher.sol, is not installed. Try installing it using npm`. I then installed solmate with yarn. Ran yarn compile again and had a similar issue with another of the submodules that the UR repo uses. Here is the error: `Error HH411: The library permit2, imported from @uniswap/universal-router/contracts/base/Dispatcher.sol, is not installed. Try installing it using npm` a Permit2 lib does not exist. 

Am I taking the right approach here? I feel like the submodules in the UR repo need to be updated to get past these errors, but am not sure how as I am using as a NPM package. Thanks!

yarn -v
`3.2.1`
node -v
`16.10.0`

marktoda:
Thanks for reporting this! Forge/hardhat interop causing issues here as we need submodules internally to universal-router for solmate and permit2. I think the best bet is to add a postinstall script to universal-router that'll handle this automatically when you yarn install. We'll try to get something out fixing that today

codeislight1:
running fixed it:
git submodule update --init --recursive

PowVT:
Thanks all really appriciate the help. Still having the same issue. @codeislight1 this command produces and error which I screen shot below. 

![Screen Shot 2022-11-23 at 8 07 44 AM](https://user-images.githubusercontent.com/21206705/203554468-9288ae96-66a2-41c8-95b7-8ca47b87db07.png)

I have made sure there is a ssh key pair generated with this command `ssh-keygen -t ed25519 -C "myemail@xyz.com"`. And get the same error. Is this a permissions thing? thanks

codeislight1:
i had a similar issue, you may fix it by downloading permit2 repo and
extract it within peemit2 folder, i believe that fixed it for me.

PowVT:
Ya it worked for the Permit2 stuff, then starts erroring on the open-zeppelin submodule. So I did the same for that, extracted in to the OZ folder in the top level directory. Then I run `git submodule update --init --recursive` again and it gets past the permit2 and OZ stuff, but errors on the solmate submodule. Do I make a new folder for solmate similar to the other 2? Do this look like your file tree? Something feels off.

[...]

PR #188, opened 2022-11-23
Simplify dependency structure for easier integration of universal-router package

ewilz:
1. Use node modules where possible.
2. Eliminate necessity for integrators to have `remappings.txt`
3. Include instructions for setting up submodules

If possible I'd like to eliminate the remappings that remove `src` directory from submodules. This could clash with other integrators who don't remap (unless of course this is the most common pattern?) cc @marktoda
~Right now this is broken since Permit2 also uses this remapping~ Made a branch https://github.com/Uniswap/permit2/pull/160

Attempting at least a short term solution to this issue: https://github.com/Uniswap/universal-router/issues/186
feedback welcome!

EDIT: with this setup and instructions, I was able to successfully compile a project with universal-router dependencies


marktoda:
> unless of course this is the most common pattern

i feel that it is, but I didn't realize hardhat doesn't support it: https://github.com/NomicFoundation/hardhat/issues/3231. So yeah I think best to remove from permit2 for now unfortunately

ewilz:
> > unless of course this is the most common pattern
> 
> i feel that it is, but I didn't realize hardhat doesn't support it: [NomicFoundation/hardhat#3231](https://github.com/NomicFoundation/hardhat/issues/3231). So yeah I think best to remove from permit2 for now unfortunately

who am I to judge as the submodules n00b, but it seems overly complicated to me to be altering file structures for imports  🤷‍♀️

marktoda:
Yea it's a builtin solc thing not a foundry thing but reasonable to just skip it if package intended to be used as a dep

.github/workflows/forge.yml
```diff
@@ -22,6 +22,28 @@ jobs:
           submodules: recursive
           token: '${{ secrets.PERMIT_POST_KEY }}'
```
marktoda, line 23: can delete now as its public no?

.github/workflows/forge.yml
```diff
@@ -22,6 +22,28 @@ jobs:
...
+      - uses: actions/cache@v1
+        with:
+          path: ${{ steps.yarn-cache.outputs.dir }}
+          key: yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            yarn-
+
+      - name: Install dependencies
+        run: |
+          npm config set '//registry.npmjs.org/:_authToken' "${{ secrets.NPM_AUTH_TOKEN }}" \
```
marktoda, line 43: do we need this? all packages should be public

marktoda (approved):
LGTM w/ couple workflow cleanup ideas

ewilz:
changed permit2 to not remap the solmate file structure, or else every dependent has to do that too

PR #189, opened 2022-11-28
[to consider] dirty reentrancy lock

hensha256:
Gas is better for our contracts with a dirty lock not a clean one - you can see this in the 2 commits


snreynolds (approved):
lgtm
just maybe worth adding a test with a malicious nft contract that tries to re-enter?

snreynolds (approved):

PR #195, opened 2022-12-07
Fix typos and misleading comments

CallMeGwei:
This commit fixes a few typos and misleading comments.


hensha256 (approved):

hensha256:
Ohhhh i think this might be failing because fork PRs can't access our github secrets 🤔 
@ewilz do you think thats right? I think it cant get the infura api key

PR #198, opened 2022-12-09
1.2.2

ewilz:
includes "speedup build process" commit for the sdk to take advantage of as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "publishConfig": {
     "access": "public"
   },
-  "version": "1.2.1",
+  "version": "1.2.2",
   "keywords": [
     "uniswap",
     "router",


ewilz:
was having some submodule dependency issues with universal-router-sdk. Turns out updating submodules correctly has more to do with committing the changes first. sdk was running forge install from inside this UR node module on `yarn install:ur` (for some reason) which would keep switching forge-std back to previous versions. This update was probably unnecessary, but here we are.

ewilz:
was having some submodule dependency issues with universal-router-sdk. Turns out updating submodules correctly has more to do with committing the changes first. sdk was running forge install from inside this UR node module on `yarn install:ur` (for some reason) which would keep switching forge-std back to previous versions. This update was probably unnecessary, but here we are.

hensha256 (approved):

PR #203, opened 2022-12-20
Increase # of commands

hensha256:


contracts/base/Dispatcher.sol
```diff
@@ -33,134 +33,148 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks
...
+                    }
+                    // 0x08 <= command < 0x10
+                } else {
+                    if (command == Commands.V2_SWAP_EXACT_IN) {
+                        (
+                            address recipient,
+                            uint256 amountIn,
+                            uint256 amountOutMin,
+                            address[] memory path,
+                            bool payerIsUser
```
hensha256, line 82: forge must have changed their linter again 🙄

ewilz:
so this allows 5 more commands to be added?

Also, I think the README table should be updated with the new command mask

ewilz (approved):
Approved pending README update.

this thing --v & if the larger table needs an update

```
 0 1 2 3 4 5 6 7
┌─┬───┬─────────┐
│f│ r | command │
└─┴───┴─────────┘
```

hensha256:
> so this allows 5 more commands to be added in the "nest"?

No it actually allows for a whole other bit to be used - so the number of commands increases from 32 to 64! Those 5 placeholders are just the 5 currently unused in the original 32

hensha256:
> fwiw I think I'm still team no nesting, but not gonna die on that hill obviously

happy to do another gas comparison but i think as we add in sells the gas savings get bigger and bigger because we will now have like 40+ commands

PR #205, opened 2023-01-04
WIP: map recipient for PERMIT2_TRANSFER_FROM

ewilz:


ewilz:
Just realized the sdk has all the address needs for transfering WETH to the router, but I think it's nice to be consistent and use mappings here to save calldata. could address the batch case too.

test/integration-tests/gas-tests/UniversalRouter.gas.test.ts
```diff
@@ -27,12 +29,13 @@ import { BigNumber } from 'ethers'
 
 const { ethers } = hre
 
-describe('UniversalRouter Gas Tests', () => {
+describe.only('UniversalRouter Gas Tests', () => {
```
hensha256, line 32:
```suggestion
describe('UniversalRouter Gas Tests', () => {
```

ewilz: womp

hensha256 (approved):

PR #208, opened 2023-01-11
feature/sell seaport

zhongeric:
Support instant selling for NFTs through Seaport.

This supports:
- Filling open orders (bids / offers), ERC721 and ERC1155 for ERC20 (WETH)

Additional notes:
- Collection & trait offers are not supported as they are not exposed in the OpenSea API yet
- For Seaport tests, now, you can specify an additional argument `itemTypes` to `getAdvancedOrderParams` as an array specifying which types of items to consider in the summation. Default is all.
- ERC721 -> ETH or ERC1155 -> ETH orders are not supported as we cannot pull the eth from the offerer (https://docs.opensea.io/v2.0/reference/seaport-overview).
  - note sells to native ETH will be supported in 1.2 per 0age 

For ERC721 -> ERC20 orders that specify additional consideration items besides just the NFT (royalites, fees), it appears that the router must be the sender of these consideration items. This would require users to append a SWEEP ERC20 or ETH command right after the sell.

Additionally, it would require the router to have approved the Seaport contract / specified Conduit contract to spend the tokens that will be received. See how the msg.sender of the call to `fufillAdvancedOrder` is the one who transfers the consideration items: https://github.com/ProjectOpenSea/seaport/blob/c30dd90272609677606f03f9c885466f15e891eb/reference/lib/ReferenceOrderFulfiller.sol#L271

Right now the flow for a NFT -> WETH order is:
- Alice approves router to spend her NFT
- Alice sends the NFT to the router (TODO: replace with Permit2 transfer when available)
- Router approves Conduit to spend its WETH (note that this approval only needs to be done once per ERC20, per conduit. If we only use OpenSea's conduit then no gas overhead here).
- Router executes sell, receiving offered WETH, sending the NFT and other consideration items
- Alice sweeps WETH back to themselves

Please lmk thoughts on this flow!


contracts/base/Dispatcher.sol
```diff
@@ -197,6 +209,27 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks
...
+        (bytes memory data, address token, address spender, uint256 id, uint256 amount, address returnAddress) =
+            abi.decode(inputs, (bytes, address, address, uint256, uint256, address));
+
+        ERC1155(token).setApprovalForAll(spender, true);
+
+        (success, output) = protocol.call(data);
+
+        // Revoke approval
+        // TODO: is this correct?
+        ERC1155(token).setApprovalForAll(spender, false);
```
marktoda, line 225: I _think_ we can just leave the approval, given the router is never meant to hold standing balance of anything

zhongeric: Alright, not sure how much gas overhead this is too

test/integration-tests/Seaport.test.ts
```diff
@@ -141,4 +157,187 @@ describe('Seaport', () => {
       router['execute(bytes,bytes[],uint256)'](commands, inputs, DEADLINE, { value: seaportValue })
     ).to.be.revertedWith('ExecutionFailed(0, "0x8baa579f")')
   })
+
+  describe('Seaport SELL 721', async () => {
+    let tubbyCats: ERC721
```
ewilz, line 162: opinionated nit (don't really need to address), if we did just keep using covens we wouldn't have to make more params for more nfts, might feel neater.

test/integration-tests/Seaport.test.ts
```diff
@@ -141,4 +157,187 @@ describe('Seaport', () => {
...
+      alice = await ethers.getSigner(ALICE_ADDRESS)
+      permit2 = (await deployPermit2()).connect(alice) as Permit2
+      router = (await deployUniversalRouter(permit2)).connect(alice) as UniversalRouter
+      planner = new RoutePlanner()
+      tubbyCats = new ethers.Contract(TUBBY_ADDRESS, ERC721_ABI) as ERC721
+      weth = new ethers.Contract(WETH.address, ERC20_ABI) as ERC20
+
+      // send 1 eth from alice to router.address for weth approval
+      await alice.sendTransaction({ to: router.address, value: ethers.utils.parseEther('1.0') })
+      expect(await ethers.provider.getBalance(router.address)).to.eq(ethers.utils.parseEther('1.0'))
```
ewilz, line 180: are there other instances of expect statements in beforeEach's in this repo??? Feels redundant... 
if it's bc there's no error if this reverts, could try 
```
(await alice.sendTransaction({ to: router.address, value: ethers.utils.parseEther('1.0') })).wait()
```

test/integration-tests/shared/protocolHelpers/seaport.ts
```diff
@@ -57,7 +79,11 @@ export function getOrderParams(apiOrder: any): { order: Order; value: BigNumber
   return { order, value }
 }
 
-export function getAdvancedOrderParams(apiOrder: any): { advancedOrder: AdvancedOrder; value: BigNumber } {
```
ewilz, line 60: what do you think about still returning `value` from this function, and having the item type filter be an optional parameter?

contracts/base/Dispatcher.sol
marktoda: double checking w/ @hensha256 @ewilz

test/integration-tests/Seaport.test.ts
```diff
@@ -141,4 +157,187 @@ describe('Seaport', () => {
...
+        tubbyCats.address,
+        OPENSEA_CONDUIT,
+        id,
+        alice.address,
+      ])
+      planner.addCommand(CommandType.SWEEP, [WETH.address, MSG_SENDER, 0])
+      const { commands, inputs } = planner
+
+      const wethBefore = await weth.connect(alice).balanceOf(alice.address)
+      expect(wethBefore).to.eq(0)
```
ewilz, line 227: I'm personally not a fan of having expect statements scattered everywhere (though I see it's happened in some files that I haven't worked on :P)

If they must be scattered, I'd rather them be before and after the call we're testing. Ideally EVERY test is testing ONE CALL ONLY. Maybe it's fine to have a list of expects before the call, and a list after. I usually put all my expectations at the bottom of the test so the SPEC IS EXTRMELY CLEAR TO THE READER and they don't have to look all over the file/test. This specific expect doesn't look necessary as later on we're doing `expect(wethAfter.sub(wethBefore)).to.eq(wethReceived)` so why does it need to be 0 to begin with?

test/integration-tests/Seaport.test.ts
```diff
@@ -141,4 +157,187 @@ describe('Seaport', () => {
...
+      // put NFT in the router TODO replace with Permit2 721
+      await tubbyCats.transferFrom(alice.address, router.address, id)
+      expect(await tubbyCats.ownerOf(id)).to.eq(router.address)
```
ewilz, line 230: also don't feel this is necessary, line 229 should just throw an errror if something goes wrong

test/integration-tests/Seaport.test.ts
```diff
@@ -141,4 +157,187 @@ describe('Seaport', () => {
...
+      await resetFork(16385143) // txn is block 16385144
+      await hre.network.provider.request({
+        method: 'hardhat_impersonateAccount',
+        params: [ALICE_ADDRESS],
+      })
+      alice = await ethers.getSigner(ALICE_ADDRESS)
+      permit2 = (await deployPermit2()).connect(alice) as Permit2
+      router = (await deployUniversalRouter(permit2)).connect(alice) as UniversalRouter
+      planner = new RoutePlanner()
+      tubbyCats = new ethers.Contract(TUBBY_ADDRESS, ERC721_ABI) as ERC721
```
ewilz, line 175: if tubbycats has to be connected with alice, can do it here so it's one and done instead of having to keep connecting her in each test.

test/integration-tests/Seaport.test.ts
```diff
@@ -141,4 +157,187 @@ describe('Seaport', () => {
...
+      const prevTubbyCatOwner = await tubbyCats.connect(alice).ownerOf(id)
+      await tubbyCats
+        .connect(await ethers.getImpersonatedSigner(prevTubbyCatOwner))
+        .transferFrom(prevTubbyCatOwner, alice.address, id)
+      tubbyCats = tubbyCats.connect(alice)
+      const ownerBefore = await tubbyCats.ownerOf(id)
+      expect(ownerBefore).to.eq(alice.address)
+
+      // Signature is invalid
+      advancedOrder.signature = '0xdeadbeef'
```
ewilz, line 261: nit: neater to do this when defining advanced order, so its instantiation is happening all in one place. might do
```
let { advancedOrder: invalidOrder, criteriaResolvers } = getAdvancedOrderParams(seaportOrders[2])
invalidOrder.signature = '0xdeadbeef'
```

test/integration-tests/Seaport.test.ts
```diff
@@ -141,4 +157,187 @@ describe('Seaport', () => {
...
+      let { advancedOrder, criteriaResolvers } = getAdvancedOrderParams(seaportOrders[2])
+      const id = criteriaResolvers[0].identifier
+      // transfer nft to alice as tubbyCatOwner
+      const prevTubbyCatOwner = await tubbyCats.connect(alice).ownerOf(id)
+      await tubbyCats
+        .connect(await ethers.getImpersonatedSigner(prevTubbyCatOwner))
+        .transferFrom(prevTubbyCatOwner, alice.address, id)
+      tubbyCats = tubbyCats.connect(alice)
+      const ownerBefore = await tubbyCats.ownerOf(id)
+      expect(ownerBefore).to.eq(alice.address)
```
ewilz, line 308: I'd get rid of 307/308, we explicitly did this transfer, and should know it worked or else the tx should error

test/integration-tests/Seaport.test.ts
```diff
@@ -141,4 +157,187 @@ describe('Seaport', () => {
...
+
+      // Signature is invalid
+      advancedOrder.signature = '0xdeadbeef'
+
+      const calldata = seaportInterface.encodeFunctionData('fulfillAdvancedOrder', [
```
ewilz, line 313: In addition to advancedOrder I'd also do calldata setup by advanced order since they'r relevant to each other instead of having tubbycat owner assignment in between which is a separate. neat organized tests are a life saver to maintain.

contracts/base/Dispatcher.sol
```diff
@@ -188,6 +200,7 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks
             abi.decode(inputs, (bytes, address, address, uint256, address));
 
         // this approval is auto-wiped when the token is transferred
+        // TODO: can we save gas here by not approving if router already owns the NFT? i.e. ERC20 -> NFT -> ETH
```
ewilz, line 203: confused by this, we're approving here so that an external protocol can pull the NFT from us (the router), so I would think it's necessary

test/integration-tests/Seaport.test.ts
zhongeric: Yeah, I think I'll port over to covens, I was just trying to find a quick POC txn. Would also help cut down on params for sure

test/integration-tests/Seaport.test.ts
zhongeric: Sg, I'll use the .wait more to get errors to throw

test/integration-tests/Seaport.test.ts
zhongeric: true, I may have gotten carried away w using expects for debugging

test/integration-tests/Seaport.test.ts
zhongeric: Going to fix after refactor to covens

test/integration-tests/Seaport.test.ts
zhongeric: Great suggestion ty!

contracts/base/Dispatcher.sol
zhongeric: Yeah hmm the comment doesnt make sense, not sure what I was referring to.

test/integration-tests/shared/protocolHelpers/seaport.ts
zhongeric: I would be fine with that, I need to remove `criteriaResolvers` as well as I don't think that's in the api response as well

contracts/base/Dispatcher.sol
ewilz: as long as we're always calling Permit2 to transfer tokens from the msg.sender, I think it should be fine to have "hanging approvals". I was trying hard a few months ago to try and think of vulnerabilities for hanging approvals, but couldn't think of any, but I do think we should give some extra scrutiny to the topic and make a full fledged decision that applies to all token types eventually. Fine to have hanging approvals for now as a default...no sure what @hensha256 thinks

test/integration-tests/Seaport.test.ts
zhongeric: Removed the 0 check, and yeah I think I'm going to refactor to use covens and only use a specific order / txn so I can move all of the nft approval stuff into the beforeEach, and only have the seaport call in the main test body

contracts/base/Dispatcher.sol
hensha256: yeah i cant really think of many issues from hanging approvals so should be okay. Though once its been approved once it then is a waste of user for other users to have line 220 at all. So maybe we should have a separate approve721 command, that the first user to sell an NFT includes in their tx?

test/integration-tests/Seaport.test.ts
```diff
@@ -141,4 +157,187 @@ describe('Seaport', () => {
...
+      weth = new ethers.Contract(WETH.address, ERC20_ABI) as ERC20
+
+      // send 1 eth from alice to router.address for weth approval
+      await alice.sendTransaction({ to: router.address, value: ethers.utils.parseEther('1.0') })
+      expect(await ethers.provider.getBalance(router.address)).to.eq(ethers.utils.parseEther('1.0'))
+      // max approve conduit for weth
+      const routerSigner = await ethers.getImpersonatedSigner(router.address)
+      await weth.connect(routerSigner).approve(OPENSEA_CONDUIT, ethers.constants.MaxUint256)
+      await weth.connect(routerSigner).allowance(router.address, OPENSEA_CONDUIT)
+      expect(await weth.connect(routerSigner).balanceOf(router.address)).to.eq(0)
```
hensha256, line 185: You can do the connect when you create weth so save you having to do connect on all of these lines :)

i.e.
```
weth = new ethers.Contract(WETH.address, ERC20_ABI).connect(routerSigner) as ERC20
```

test/integration-tests/Seaport.test.ts
hensha256: I agree with emily on the point that I would like all the expects gathered into 2 blocks
- expects immediately before the router call
- expects immediately after the router call

In general I prefer to err on the side of more instead of less (differently to emily). If adding an expect means you are more certain that the expected behaviour is occurring then I'd say add it - it increases the reliability and security of your tests.

However I do agree that we dont actually care _what_ alice's balance is before the test - we just care that it goes up by the expected amount.

test/integration-tests/Seaport.test.ts
```diff
@@ -141,4 +157,187 @@ describe('Seaport', () => {
...
+
+    it('reverts if order does not go through', async () => {
+      let { advancedOrder, criteriaResolvers } = getAdvancedOrderParams(seaportOrders[2])
+      const id = criteriaResolvers[0].identifier
+      // transfer nft to alice as tubbyCatOwner
+      const prevTubbyCatOwner = await tubbyCats.connect(alice).ownerOf(id)
+      await tubbyCats
+        .connect(await ethers.getImpersonatedSigner(prevTubbyCatOwner))
+        .transferFrom(prevTubbyCatOwner, alice.address, id)
+      tubbyCats = tubbyCats.connect(alice)
```
hensha256, line 306: you should move this somewhere else - so that you can remove the additional connection from line 302

contracts/base/Dispatcher.sol
zhongeric: Do you mean approve1155? And that once the router approves Seaport for the entire 1155 collection we shouldn't have another approval? If we want to do hanging approvals we could do something like 
```solidity
if(!isApprovedForAll(address(this), spender) {
 ERC1155(token).setApprovalForAll(spender, true);
}
```

test/integration-tests/Seaport.test.ts
zhongeric: Sounds good, changed in latest commit to have all of the expects at the very end of the test case, and `before` vars strictly before the call to the router, and all `after` vars strictly after.

contracts/base/Dispatcher.sol
ewilz: @hensha256 means that `approve` and `sell` are 2 completely different commands. You could actually imagine collection creators triggering this to save their traders gas. Interesting idea, and a nice-to-have depending on gas savings.

test/integration-tests/Seaport.test.ts
ewilz: I don't think I was saying that less expects are better in general. Just that there shouldn't be expects testing things other than the call we're testing.

contracts/base/Dispatcher.sol
zhongeric: Ah I see, yeah that makes sense to me. I'd support decoupling the approve from the selling command - however this would only make sense for 1155 right? Since for 721 approval is cleared on transfer

test/integration-tests/shared/protocolHelpers/seaport.ts
```diff
@@ -47,6 +56,19 @@ export type AdvancedOrder = Order & {
...
+  CONSIDERATION,
+}
+
+export type CriteriaResovler = {
+  rderIndex: BigNumber
+  side: Side // enum
+  index: BigNumber
+  identifier: BigNumber
+  criteriaProof: string[] // bytes32[]
+}
```
zhongeric, line 70: Not used now, but can leave in for the future when criteria orders are supported by OS api

hensha256:
Closing for now but the branch will remain

PR #212, opened 2023-01-20
Batch reverts

hensha256:
- Recipient mapping library and ReentrancyLock contract merged so that sub-calls can still access the outermost `msg.sender`


contracts/base/LockAndMsgSender.sol
```diff
@@ -3,15 +3,29 @@ pragma solidity ^0.8.17;
...
+    address private constant ADDRESS_ONE = address(1);
+    address internal isLockedSender = ADDRESS_ONE;
+
+    modifier isNotLocked() {
+        if (msg.sender != address(this)) {
+            if (isLockedSender != ADDRESS_ONE) revert ContractLocked();
+            isLockedSender = msg.sender;
+            _;
+            isLockedSender = ADDRESS_ONE;
+        } else {
```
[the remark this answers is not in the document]
hensha256: Yes all reentrancy with `msg.sender==address(this)` would be allowed with this. However:

- only one function has the reentrancy lock so its only 1 entry point to consider
- all of the `.call(` operations in the codebase are performed upon a fixed address - mostly protocols but in 1 place it is `(address(this)).call`. This is _only_ within the `EXECUTE_SUB_PLAN` command which is defined in this PR.

So yes it does loosen the reentrancy lock a bit but its such a tight set of circumstances that it feels safe to me.

However, interested to know any concerns you have!

contracts/base/LockAndMsgSender.sol
```diff
@@ -3,15 +3,29 @@ pragma solidity ^0.8.17;
...
 import {Constants} from '../libraries/Constants.sol';
 
-/// @title Recipient Library
-/// @notice Calculates the recipient address for a command
-library Recipient {
+contract LockAndMsgSender {
+    error ContractLocked();
+
+    address private constant ADDRESS_ONE = address(1);
+    address internal isLockedSender = ADDRESS_ONE;
```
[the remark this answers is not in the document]
hensha256: yeah good point. Renaming to `lockedBy` but open to suggestions

contracts/UniversalRouter.sol
```diff
@@ -55,6 +76,198 @@ contract UniversalRouter is RouterImmutables, IUniversalRouter, Dispatcher, Rewa
         return command & Commands.FLAG_ALLOW_REVERT == 0;
     }
 
-    // To receive ETH from WETH and NFT protocols
+    /// @notice Decodes and executes the given command with the given inputs
```
ewilz, line 79: wait! I kind of miss having a separate dispatcher, felt so clean. why remove it?

ewilz: can it do `address(this).call(data)` ?

contracts/base/LockAndMsgSender.sol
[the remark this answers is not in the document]
ewilz: Yah the only wayy ever this can be reentered is if the UR itself calls execute(). I don't think there's any way for a malicious token for instance, to get the UR to do this without breaking this lock, must be signed off by the original caller.

contracts/UniversalRouter.sol
hensha256: yeah it can - thats what it already does if you look 😆 
but my point is that it breaks nice coding design if the parent contract is calling a function on the child contract. The parent contract should be designed in a way that does not rely on the design of the inheriting child contract.

snreynolds: But you can still have the contracts be separated and use `address(this).call(data)`. Alternatively could you define an execute function inside the Dispatcher.sol contract that is virtual and in UniversalRouter just override it? 

I do think the separation of the contracts is pretty clean, but will leave it up to you guys to make the final call.

hensha256: > But you can still have the contracts be separated and use address(this).call(data). 

Yes sorry - I agree that this is _possible_, to me its just a bad programming style to have a contract call a function that it doesnt even know will ever exist

> Alternatively could you define an execute function inside the Dispatcher.sol contract that is virtual and in UniversalRouter just override it

This however is a much better solution because its then a function that the Dispatcher knows is guaranteed to exist.

contracts/base/Dispatcher.sol
```diff
@@ -316,11 +315,22 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks
...
+                // placeholder area for commands 0x21-0x3f
+                revert InvalidCommandType(command);
+            }
         }
     }
 
+    /// @notice Executes encoded commands along with provided inputs.
+    /// @param commands A set of concatenated commands, each 1 byte in length
+    /// @param inputs An array of byte strings containing abi encoded inputs for each command
+    function execute(bytes calldata commands, bytes[] calldata inputs) external payable virtual;
```
ewilz, line 332: kind of awkward. If I was coding it, I'd probably just allow the dispatcher to arbitrarily call back into itself like we call the protocols but maybe being more rigid is the right move despite the awkwardness?

hensha256: This function actually doesnt _have_ to be here because we do the call to `execute` as a low level call. However that low level call does of course mean that all inheriting contracts must implement an execute function - which is why i put this here.

PR #216, opened 2023-01-26
Wrapping optimisation

hensha256:


test/integration-tests/gas-tests/__snapshots__/Uniswap.gas.test.ts.snap
```diff
@@ -45,35 +45,35 @@ Object {
...
-  "gasUsed": 190555,
+  "gasUsed": 190554,
 }
 `;
 
 exports[`Uniswap Gas Tests Mixing V2 and V3 with Universal Router. Split routes gas: ETH --> ERC20 split V2 and V3, one hop 1`] = `
 Object {
   "calldataByteLength": 1124,
-  "gasUsed": 218582,
+  "gasUsed": 204192,
```
hensha256, line 76: !!!

contracts/modules/Payments.sol
```diff
@@ -101,7 +101,9 @@ abstract contract Payments is RouterImmutables {
         }
         if (amount > 0) {
             WETH9.deposit{value: amount}();
-            WETH9.transfer(recipient, amount);
+            if (recipient != address(this)) {
```
NoahZinsmeister, line 104: interesting, i'm struggling to think of a path where this would save gas - is there a case where we can't wrap ETH + transfer WETH from within a v3 callback?

hensha256: we chatted on slack - but basically this is because we separate out wrap/unwrap from trading so that the command can be used for NFT trades and other situations too. Also useful for when we have split trades on uniswap to only have to do a single wrap

marktoda (approved):
nice seems reasonable. weth deposit/withdraw should take recipient param :upside_down_face:

PR #209, opened 2023-01-17
Prefer using calldata instead of memory

k06a:
Sometimes it's a lot cheaper in terms of gas, sometimes it's a bit.

I was made to decrease `runs` setting in hardhat config due slightly increased bytecode size (out of 24kb limit).

PR #218, opened 2023-01-30
Prefer using calldata instead of memory

hensha256:
Local branch PR from @k06a 's PR #209 

* Prefer using calldata instead of memory
* Extensive calldata modifier usage over the source code
* Fix Solidity compiler options in Foundry config

Additional changes from me

* Addition of `toPool` and removal of `toUint24`
* Removal of now uneccessary length checks


contracts/modules/uniswap/v3/V3Path.sol
```diff
@@ -44,17 +53,19 @@ library V3Path {
...
     }
 
-    /// @notice Skips a token + fee element from the buffer in place
+    /// @notice Skips a token + fee element
     /// @param path The swap path
-    function skipToken(bytes memory path) internal pure {
-        path.inPlaceSliceToken(path.length - NEXT_OFFSET);
+    function skipToken(bytes calldata path) internal pure returns (bytes calldata) {
+        if (path.length < ADDR_AND_FEE_LENGTH) revert SliceOutOfBounds();
+        return path[ADDR_AND_FEE_LENGTH:];
```
NoahZinsmeister, line 69: prefer `NEXT_OFFSET`, then we can delete `ADDR_AND_FEE_LENGTH`

contracts/modules/uniswap/v3/V3Path.sol
```diff
@@ -44,17 +53,19 @@ library V3Path {
     /// @notice Gets the segment corresponding to the first pool in the path
     /// @param path The bytes encoded swap path
     /// @return The segment containing all data necessary to target the first pool in the path
-    function getFirstPool(bytes memory path) internal pure returns (bytes memory) {
-        return path.slicePool();
+    function getFirstPool(bytes calldata path) internal pure returns (bytes calldata) {
+        if (path.length < POOL_LENGTH) revert SliceOutOfBounds();
+        return path[:POOL_LENGTH];
```
NoahZinsmeister, line 58: prefer `POP_OFFSET`, then we can delete `POOL_LENGTH`

NoahZinsmeister: also feel free to rename for clarity

contracts/modules/uniswap/v3/V3Path.sol
NoahZinsmeister: also feel free to rename for clarity

contracts/modules/uniswap/v3/BytesLib.sol
```diff
@@ -128,7 +29,7 @@ library BytesLib {
         }
 
         assembly {
-            tempAddress := mload(add(add(_bytes, 0x14), _start))
+            tempAddress := shr(96, calldataload(add(_bytes.offset, _start)))
```
NoahZinsmeister, line 32: prefer all hex or all decimal in assembly blocks

NoahZinsmeister (commented):
looks pretty good on first glance

k06a:
BTW, we are requesting Solidity compiler to allow `abi.decode()` to process `calldata` to `calldata`: https://github.com/ethereum/solidity/issues/13518

hensha256:
Pretty sure #219 will fix the issues we're having with hardhat here

foundry.toml
```diff
@@ -4,7 +4,7 @@ out = 'out'
 libs = ['lib']
 via_ir = true
 solc_version = '0.8.17'
-optimizer_runs = 1_000_000
+optimizer_runs = 10_000
```
marktoda, line 7: were you getting better results with this?

contracts/base/Dispatcher.sol
```diff
@@ -79,24 +84,32 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks
...
+                        address[] calldata path = inputs.toAddressArray(3);
                         address payer = payerIsUser ? msg.sender : address(this);
                         v2SwapExactInput(recipient.map(), amountIn, amountOutMin, path, payer);
                     } else if (command == Commands.V2_SWAP_EXACT_OUT) {
                         (
                             address recipient,
                             uint256 amountOut,
                             uint256 amountInMax,
-                            address[] memory path,
+                            /* address[] memory path */
```
marktoda, line 99: delete?

marktoda (approved):
seems reasonable to me

foundry.toml
hensha256: its because bytecode size went up so much we could no longer have 1M runs 😢 will look into getting bytecode down next week so we can put this back up

contracts/base/Dispatcher.sol
hensha256: i think we should keep it for readablility because theres 1 element not being decoded. But going to change the comment formatting so its clearer

marktoda (approved):

PR #224, opened 2023-02-06
Add support for Seaport 1.2

zhongeric:
TLDR: For our use case, Seaport1.2 is basically the same as Seaport1.1

Most of the changes are the addition of the new ABI, gas tests (expected bc new deployment addr added)

When OpenSea moves their interface over to 1.2, we can update the test in the repo to use a better order - right now its using someone's order they made directly to Seaport (which still works just fine just doesn't use the OpenSea conduit).


contracts/base/Dispatcher.sol
```diff
@@ -181,8 +181,12 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks
                         (address token, address recipient, uint256 id, uint256 amount) =
                             abi.decode(inputs, (address, address, uint256, uint256));
                         Payments.sweepERC1155(token, recipient.map(), id, amount);
+                    } else if (command == Commands.SEAPORT_V2) {
```
ewilz, line 184: any reason to keep support for the previous Seaport in?

zhongeric: I think while they plan on using the same conduit for Seaport1.2 as the previous one, 0age mentioned they will be removing the old version 1.1 in the future:
```
If you're currently building on v1.1, you're welcome to keep using it (though it will eventually be removed from 
[@OpenSea](https://twitter.com/opensea)
's conduit)
```
https://twitter.com/z0age/status/1621240001114836992

test/integration-tests/Seaport.test.ts
```diff
@@ -145,3 +147,113 @@ describe('Seaport', () => {
       .withArgs(0, customErrorSelector)
   })
 })
+
+describe('SeaportV2', () => {
```
zhongeric, line 151: Optionally, we could reduce code redundancy (right now its basically copy paste from the original test suite) by looping over the original one with the new Commands.SEAPORT_V2 and the new NFT contract address. However the test where we buy two cryptoCovens for `fulfillAvailableAdvancedOrders` wouldn't work as we don't have the orders right now.

contracts/base/Dispatcher.sol
```diff
@@ -308,8 +308,12 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks
                             amount := calldataload(add(inputs.offset, 0x60))
                         }
                         Payments.sweepERC1155(token, recipient.map(), id, amount);
+                    } else if (command == Commands.SEAPORT_V2) {
+                        (uint256 value,) = abi.decode(inputs, (uint256, bytes));
```
hensha256, line 312: for gas reasons I now decode in assembly for the other commands... its kind of ugly but more efficient. Look at how the other SEAPORT command is decoded to see what I mean

contracts/libraries/Commands.sol
```diff
@@ -47,5 +47,8 @@ library Commands {
     uint256 constant X2Y2_1155 = 0x1b;
     uint256 constant FOUNDATION = 0x1c;
     uint256 constant SWEEP_ERC1155 = 0x1d;
-    // COMMAND_PLACEHOLDER = 0x1e to 0x3f (all unused)
+    uint256 constant SEAPORT_V2 = 0x1e;
```
hensha256, line 50: given seaport is our most used NFT marketplace I think we should maybe change this to be 0x21 (I'm about to merge a PR that takes 0x20). That way it has minimal `if` checks before executing

contracts/base/Dispatcher.sol
```diff
@@ -308,8 +308,12 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks
...
+                        bytes calldata data = inputs.toBytes(1);
+                        (success, output) = SEAPORT_V2.call{value: value}(data);
```
hensha256, line 314: I think we should have a very clear comment here and in the readme about the fact that this contract is _not_ secure if the `offerer` of the order is a contract, and a very quick high level sentence explaining the attack, and a link to their doc that talks about it

contracts/base/Dispatcher.sol
zhongeric: Ah yes I see, I'll change

contracts/base/Dispatcher.sol
zhongeric:
```
                /// @dev Seaport 1.2 allows for orders to be created by contracts. 
                ///     These orders pass control to the contract offerers during fufillment,
                ///         allowing them to perform any number of destructive actions as a holder of the NFT. 
                ///     Integrators should be aware that in some scenarios: e.g. purchasing an NFT that allows the holder
                ///         to claim another NFT, the contract offerer can "steal" the claim during order fufillment. 
                ///     This is also outlined in the Seaport documentation: https://github.com/ProjectOpenSea/seaport/blob/main/docs/SeaportDocumentation.md
```

contracts/base/Dispatcher.sol
```diff
@@ -308,7 +308,7 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
                         }
                         Payments.sweepERC1155(token, map(recipient), id, amount);
                     } else {
-                        // placeholder area for commands 0x1e-0x1f
+                        // placeholder area for command 0x1e-0x1f
```
hensha256, line 311:
```suggestion
                        // placeholder area for commands 0x1e-0x1f
```

contracts/libraries/Commands.sol
```diff
@@ -8,7 +8,7 @@ library Commands {
     bytes1 internal constant FLAG_ALLOW_REVERT = 0x80;
     bytes1 internal constant COMMAND_TYPE_MASK = 0x3f;
 
-    // Command Types. Maximum supported command at this moment is 0x1F.
+    // Command Types. Maximum supported command at this moment is 0x1e.
```
hensha256, line 11:
```suggestion
    // Command Types. Maximum supported command at this moment is 0x3f.
```

test/integration-tests/shared/protocolHelpers/seaport.ts
```diff
@@ -13,6 +14,9 @@ export const seaportV2Orders = JSON.parse(
   fs.readFileSync('test/integration-tests/shared/orders/SeaportV2.json', { encoding: 'utf8' })
 )
 export const seaportInterface = new ethers.utils.Interface(SEAPORT_ABI)
+export const seaportV2Interface = new ethers.utils.Interface(SEAPORT_V2_ABI)
+// @dev 0 addr for an order that was not sent through the OpenSea conduit
```
hensha256, line 18: > 0 addr
This isnt an address - its a bytes32 right?

hensha256 (approved):

test/integration-tests/shared/protocolHelpers/seaport.ts
zhongeric: You're right! will change

PR #226, opened 2023-02-09
Add element market

zhongeric:


test/integration-tests/Element.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  let planner: RoutePlanner
+  let testNFTContract: ERC721
+
+  const order = EXAMPLE_ETH_SELL_ORDER
+  const signature = EXAMPLE_ETH_SELL_ORDER_SIG
+  const nftContractAddress = order.nft
+
+  beforeEach(async () => {
+    // txn is at block 16627214
+    await resetFork(16627214 - 1)
```
hensha256, line 31: did you try to test on polygon itself? was it too hard?
I know infura are having polygon issues this week so not worth trying now haha

test/integration-tests/Element.test.ts
```diff
@@ -0,0 +1,67 @@
...
+import { BigNumber } from 'ethers'
+import { abi as ERC721_ABI } from '../../artifacts/solmate/src/tokens/ERC721.sol/ERC721.json'
+import { expect } from 'chai'
+import { EXAMPLE_ETH_SELL_ORDER, EXAMPLE_ETH_SELL_ORDER_SIG } from './shared/protocolHelpers/element'
+
+const { ethers } = hre
+
+const ELEMENT_721_INTERFACE = new ethers.utils.Interface(ELEMENT_ABI)
+
+describe('Element Market', () => {
```
hensha256, line 18: would be great to get a gas test too!

test/integration-tests/Element.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    testNFTContract = new ethers.Contract(nftContractAddress, ERC721_ABI).connect(alice) as ERC721
+  })
+
+  it('purchases open order', async () => {
+    const value = BigNumber.from(order.erc20TokenAmount)
+    const calldata = ELEMENT_721_INTERFACE.encodeFunctionData('buyERC721Ex', [
+      order,
+      signature,
+      order.taker, // taker
+      '0x00', // extraData
```
hensha256, line 50: nit: can it just be `0x`?

test/integration-tests/Element.test.ts
zhongeric: I was running into a really weird bug where the polygon fork would only work if the default hardhat network is set to polygon - and when I did that polygon was working. But that would break all of our other tests

hensha256 (approved):

PR #234, opened 2023-02-16
Modify pay portion

hensha256:


contracts/modules/Payments.sol
```diff
@@ -45,12 +45,16 @@ abstract contract Payments is RouterImmutables {
         if (token == Constants.ETH) {
             uint256 balance = address(this).balance;
             uint256 amount = (balance * bips) / FEE_BIPS_BASE;
-            recipient.safeTransferETH(amount);
+            uint256 fixedAmount = (amount * PAYMENT_AMOUNT_BIPS) / FEE_BIPS_BASE;
```
snreynolds, line 48: do we want to handle case when this these are unset and `fixedAmount` is 0? I.e if contract is initialized with `PAYMENT_AMOUNT_BIPS` is 0 and/or `PAYMENT_RECIPIENT` is addr 0? Would save a call

hensha256: I think in that case we just wouldnt deploy with this feature. it wont be modifiable so if we want `PAYMENT_AMOUNT_BIPS==0` we would remove this code entirely

marktoda (approved):

contracts/modules/Payments.sol
```diff
@@ -61,12 +61,16 @@ abstract contract Payments is RouterImmutables {
...
```
ewilz, line 64: Our past routers have had the option to let third parties use this feature too, are we ok with getting rid of that?

hensha256: I'm confused... what about this doesnt let a third party use it?
The design is:
- x% is taken from the user
- `PAYMENT_AMOUNT_BIPS`% of the x% goes to the recipient
- the rest of the x% goes to the address specified by the 3rd party

PR #247, opened 2023-03-08
Refactor `BytesLib`

hensha256:
Internal PR from fork PR #245 so we can run CI

### DIFF
--- a/contracts/modules/uniswap/v3/BytesLib.sol
+++ b/contracts/modules/uniswap/v3/BytesLib.sol
@@ -29,45 +29,53 @@ library BytesLib {
         }
     }
 
-    /// @notice Decode the `arg`-th element in `_bytes` as a dynamic array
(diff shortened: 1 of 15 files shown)


contracts/modules/uniswap/v3/BytesLib.sol
```diff
@@ -29,27 +29,48 @@ library BytesLib {
         }
     }
 
-    function toBytes(bytes calldata _bytes, uint256 arg) internal pure returns (bytes calldata res) {
+    /// @notice Decode the `arg`-th element in `_bytes` as a dynamic array
+    /// @dev The decoding of `length` and `offset` is universal,
```
marktoda, line 33: nit add @params

marktoda (commented):
Noticing we dont have any BytesLib unit tests? probably because we were originally using the standard version verbatim... but as we make changes think it would be nice to have full unit + fuzz test coverage here for peace of mind?

PR #248, opened 2023-03-09
DeployScript bug: startBroadcast() in runAndDeployPermit2 is not stopped

zeroknots:
When using runAndDeployPermit2() in the deployment script, 
1. permit2 is first deployed and the address written into params. _(uses vm.startBroadcast() and does not end it)_
2. run() function is called and UniversalRouter deployed. (_starts a new vm.startBroadcast())_

This will cause the UniversalRouter deployment to fail.

Very simple fix, but figured I should do a pull request in case other people experience the same issue when integrating the deploy scripts into their build environment.


marktoda:
closing in favor of https://github.com/Uniswap/universal-router/pull/260 for ci

PR #260, opened 2023-03-14
Fixed broadcast() bug

marktoda:
#248

### DIFF
--- a/script/DeployUniversalRouter.s.sol
+++ b/script/DeployUniversalRouter.s.sol
@@ -60,7 +60,7 @@ contract DeployUniversalRouter is Script {
 
     function runAndDeployPermit2(string memory pathToJSON) public returns (UniversalRouter router) {
         RouterParameters memory params = fetchParameters(pathToJSON);
-        vm.startBroadcast();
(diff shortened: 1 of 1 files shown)


marktoda:
When using runAndDeployPermit2() in the deployment script,

permit2 is first deployed and the address written into params. (uses vm.startBroadcast() and does not end it)
run() function is called and UniversalRouter deployed. (starts a new vm.startBroadcast())
This will cause the UniversalRouter deployment to fail.

Very simple fix, but figured I should do a pull request in case other people experience the same issue when integrating the deploy scripts into their build environment.

marktoda:
When using runAndDeployPermit2() in the deployment script,

permit2 is first deployed and the address written into params. (uses vm.startBroadcast() and does not end it)
run() function is called and UniversalRouter deployed. (starts a new vm.startBroadcast())
This will cause the UniversalRouter deployment to fail.

Very simple fix, but figured I should do a pull request in case other people experience the same issue when integrating the deploy scripts into their build environment.

hensha256 (approved):

PR #264, opened 2023-03-20
LooksRareV2 Support

hensha256:


contracts/base/Dispatcher.sol
```diff
@@ -337,12 +337,7 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
             }
             // 0x20 <= command
         } else {
-            if (command == Commands.EXECUTE_SUB_PLAN) {
```
hensha256, line 340: just moving this further down for gas reasons

script/deployParameters/DeployGoerli.s.sol
```diff
@@ -20,6 +20,7 @@ contract DeployGoerli is DeployUniversalRouter {
             nft20Zap: UNSUPPORTED_PROTOCOL,
             cryptopunks: UNSUPPORTED_PROTOCOL,
             looksRare: 0xD112466471b5438C1ca2D218694200e49d81D047,
+            looksRareV2: UNSUPPORTED_PROTOCOL, // this is incorrect - have asked the LR team for the address
```
hensha256, line 23: * need to update when LR reply

test/integration-tests/LooksRare.test.ts
```diff
@@ -100,3 +101,52 @@ describe('LooksRare', () => {
...
+      order.affiliate,
+    ])
+    planner.addCommand(CommandType.LOOKS_RARE_V2, [value, calldata])
+
+    const { commands, inputs } = planner
+
+    await expect((await dragonNFT.connect(alice).ownerOf(tokenId)).toLowerCase()).not.to.eq(ALICE_ADDRESS)
+
+    await router['execute(bytes,bytes[],uint256)'](commands, inputs, DEADLINE, { value })
+    await expect((await dragonNFT.connect(alice).ownerOf(905)).toLowerCase()).to.eq(ALICE_ADDRESS)
```
zhongeric, line 150: nit: use `tokenId` instead of 905

zhongeric (approved):
Lgtm! Only things I'd flag are to make sure the existing logic for `looksRareRewardsDistributor` works with V2, and that the data schema of the example V2 order matches what's returned by the API.

zhongeric (approved):

contracts/base/Dispatcher.sol
```diff
@@ -356,6 +351,19 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
                 }
                 bytes calldata data = inputs.toBytes(1);
                 (success, output) = SEAPORT_V1_4.call{value: value}(data);
+            } else if (command == Commands.LOOKS_RARE_V2) {
+                // equivalent: abi.decode(inputs, (uint256, bytes))
+                uint256 value;
+                assembly {
+                    value := calldataload(inputs.offset)
```
ewilz, line 358: wondering whether this data/value thing gets repeated enough that a helper function might DRY things up?

test/integration-tests/gas-tests/LooksRare.gas.test.ts
```diff
@@ -98,3 +104,42 @@ describe('LooksRare Gas Tests', () => {
     })
   })
 })
+
+describe('LooksRareV2 Gas Test', () => {
```
ewilz, line 108: does LooksRare V2 allow bulk buys now?? Should we test that?

contracts/base/Dispatcher.sol
```diff
@@ -356,6 +351,19 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
                 }
                 bytes calldata data = inputs.toBytes(1);
                 (success, output) = SEAPORT_V1_4.call{value: value}(data);
+            } else if (command == Commands.LOOKS_RARE_V2) {
```
ewilz, line 354: Is there a point in keeping `Commands.LOOKS_RARE_721` and `Command.LOOKS_RARE_1155` around?? maybe we're afraid they might have an emergency rollback?

test/integration-tests/LooksRare.test.ts
```diff
@@ -100,3 +106,48 @@ describe('LooksRare', () => {
     })
   })
 })
+
+describe('LooksRareV2', () => {
```
ewilz, line 110: does LooksRare V2 allow bulk buys now?? Should we test that?

contracts/base/Dispatcher.sol
hensha256: yeah nice idea. doing this in #276 woo

ewilz (approved):
lgtm, not totally against keeping the old looksrare around if we're afraid they'll rollback. We don't test 1155's but since this acts a bit more like seaport now, that's more on looksrare's part so I don't think it's SUPER important we test, but couldn't hurt..

PR #269, opened 2023-03-27
Remove json from deploys main

hensha256:
See #268 
having to do 2 PRs while we're maintaining 2 versions of UR for the next month


marktoda (approved):
lgtm ref #268

PR #270, opened 2023-03-29
ABDK Audit: Lock issues

hensha256:
Issue 16 only


contracts/base/Dispatcher.sol
```diff
@@ -55,7 +55,7 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
                             payerIsUser := calldataload(add(inputs.offset, 0x80))
                         }
                         bytes calldata path = inputs.toBytes(3);
-                        address payer = payerIsUser ? lockedBy : address(this);
+                        address payer = payerIsUser ? lockedBy() : address(this);
```
ewilz, line 58: I'm confused about this fix. Isn't it impossible to get here without it being locked? It's an internal function that's only called by a function that locks the contract.

ewilz: (also issue 5 seems to be about Payments, not locks)

hensha256: issue 5 is _row_ 6 because of the header row. The issue numbers are in column E :)

snreynolds (approved):
lgtm

PR #272, opened 2023-03-30
ABDK Audit: Nits

hensha256:
Issues 15, 18, 19, and 20

### DIFF
--- a/contracts/base/Dispatcher.sol
+++ b/contracts/base/Dispatcher.sol
@@ -357,7 +357,7 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
                 bytes calldata data = inputs.toBytes(1);
                 (success, output) = SEAPORT_V1_4.call{value: value}(data);
             } else if (command == Commands.APPROVE_ERC20) {
-                address token;
+                ERC20 token;
                 uint256 spenderID;
                 assembly {
                     token := calldataload(inputs.offset)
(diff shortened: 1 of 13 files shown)


marktoda (approved):

contracts/base/Dispatcher.sol
```diff
@@ -37,10 +37,10 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
 
         success = true;
 
-        if (command < 0x20) {
-            if (command < 0x10) {
+        if (command < Commands.FOURTH_IF_BOUNDARY) {
```
ewilz, line 40: haha I find this less readable, but I guess more consistent since it's also constants.

contracts/libraries/Commands.sol
```diff
@@ -8,6 +8,13 @@ library Commands {
     bytes1 internal constant FLAG_ALLOW_REVERT = 0x80;
     bytes1 internal constant COMMAND_TYPE_MASK = 0x3f;
 
+    // The commands are executed in nested if blocks to minimise gas consumption
+    // The following constants define the boundaries where the if blocks split commands
+    uint256 constant FIRST_IF_BOUNDARY = 0x08;
+    uint256 constant SECOND_IF_BOUNDARY = 0x10;
+    uint256 constant THIRD_IF_BOUNDARY = 0x18;
+    uint256 constant FOURTH_IF_BOUNDARY = 0x20;
```
marktoda, line 16: this is kinda weird and super nit, but maybe the point of these would be more clear if we inter-sperse them in the commands below? that way its super clear exactly what we're boundary-izing. like put FIRST_IF_BOUNDARY after PAY_PORTION and second after BALANCE_CHECK Etc

marktoda (approved):

PR #273, opened 2023-04-03
ABDK Audit: Pay portion issues

hensha256:
Issue 6

Issue 6:
- Logic in `payPortion` is "suspicious" and undocumented

### DIFF
--- a/contracts/modules/Payments.sol
+++ b/contracts/modules/Payments.sol
@@ -57,18 +57,14 @@ abstract contract Payments is RouterImmutables {
     /// @param recipient The address that will receive payment
     /// @param bips Portion in bips of whole balance of the contract
     function payPortion(address token, address recipient, uint256 bips) internal {
-        if (bips == 0 || bips > 10_000) revert InvalidBips();
+        if (bips == 0 || bips > FEE_BIPS_BASE) revert InvalidBips();
         if (token == Constants.ETH) {
             uint256 balance = address(this).balance;
-            // store intermediate value to protect against phantom overflow
-            uint256 numerator = balance * bips;
-            uint256 amount = numerator / FEE_BIPS_BASE;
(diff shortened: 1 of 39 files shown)


contracts/modules/Payments.sol
```diff
@@ -60,17 +60,16 @@ abstract contract Payments is RouterImmutables {
         if (bips == 0 || bips > 10_000) revert InvalidBips();
         if (token == Constants.ETH) {
             uint256 balance = address(this).balance;
-            uint256 amount = (balance * bips) / FEE_BIPS_BASE;
-            uint256 fixedAmount = (amount * PAYMENT_AMOUNT_BIPS) / FEE_BIPS_BASE;
```
azflin, line 64: Why are you removing `fixedAmount`? What did this used to do

contracts/modules/Payments.sol
```diff
@@ -60,17 +60,16 @@ abstract contract Payments is RouterImmutables {
...
-            PAYMENT_RECIPIENT.safeTransferETH(fixedAmount);
-            recipient.safeTransferETH(amount - fixedAmount);
+            // store intermediate value to protect against phantom overflow
+            uint256 numerator = balance * bips;
```
azflin, line 64: How does storing intermediate value protect against phantom overflow?
In both below cases, revert occurs when `balance * bips` > max uint256.
```
uint256 amount = (balance * bips) / FEE_BIPS_BASE;


uint256 numerator = balance * bips;
uint256 amount = numerator / FEE_BIPS_BASE;
```

contracts/modules/Payments.sol
hensha256: it used to send a % of the fee taken to another address, but thats being removed

marktoda (approved):

PR #291, opened 2023-05-07
Gas savings

### LINKED ISSUE #290: The upper 12 bytes of `CREATE2` pool address computed are not cleaned
In `UniswapV2Library.pairForPreSorted` and `V3SwapRouter.computePoolAddress`, the pool address is computed in pure Solidity and casted via `address(uint160(uint256()))`. However the upper 12 bytes are not explicitly cleaned and the address is later used in [`solmate::SafeTransferLib.safeTransfer`](https://github.com/transmissions11/solmate/blob/2001af43aedb46fdc2335d2a7714fb2dae7cfcd1/src/utils/SafeTransferLib.sol#L79) which is written in inline assembly, where the full 32 bytes of `to` are copied to memory and directly forwarded to the ERC20 contract being called. One shouldn't assume the upper bits are cleaned when casting a `keccak256` hash to `address`.

When compiled with `via_ir` enabled, all tests pass. However, after making changes that resolve "stack too deep" errors and compiling without the IR pipeline, some Foundry tests failed in an `ERC20.transfer`. Changing `safeTransfer` back to the organic `transfer` solves the problem.

It was discovered that the dirty upper bits of a pool address are the culprit, but somehow the IR pipeline may clean the address after `keccak256`. Nonetheless, the upper 12 bytes of an address computed via `keccak256` should be cleaned explicitly.

shuhuiluo:
### Refactor `_v2Swap` to resolve stack too deep error

Previously `UniversalRouter` can only be compiled via the IR pipeline due to the "stack too deep" error which takes a long time especially in Foundry. After refactoring `_v2Swap` and modifying [`SignatureTransfer.permitWitnessTransferFrom`](https://github.com/Uniswap/permit2/blob/a5c928d037ff41b6940bf5e51ccf00e50d621bec/src/SignatureTransfer.sol#L40-L42) to
```Solidity
bytes32 hash = permit.hashWithWitness(witness, witnessTypeString);
_permitTransferFrom(permit, transferDetails, owner, hash, signature);
```
it not only saves gas but the project can also be compiled with `via_ir = false` which allows for faster compilation and testing

### Compute v2 and v3 pool address using inline assembly and explicitly clean the upper bits

Previously the pool address was computed in pure Solidity and casted via `address(uint160(uint256()))`. However the upper 12 bytes are not explicitly cleaned and the address is later used in [`solmate::SafeTransferLib.safeTransfer`](https://github.com/transmissions11/solmate/blob/2001af43aedb46fdc2335d2a7714fb2dae7cfcd1/src/utils/SafeTransferLib.sol#L79) which is written in inline assembly. When compiled with `via_ir` enabled, all tests pass. However, after making the aforementioned changes, some Foundry tests failed in a `ERC20.transfer` with `via_ir` disabled. It was discovered that the dirty upper bits of pool address are the culprit, but somehow the IR pipeline may clean the address after `keccak256`. Nonetheless, `pairForPreSorted` and `computePoolAddress` are rewritten in inline assembly to save gas and clean the upper bits. Closes #290.

### Add Uniswap v3 Foundry tests for more granular gas comparison

There weren't Uniswap v3 tests in Foundry. In order to validate gas optimizations to be made, test contracts `UniswapV3Test` and `V3MockMock` have been added. `forge snapshot --diff` is run after each modification to validate the gas savings.

### Replace conditional statements with bitwise operations

The current Solidity optimizer isn't smart enough to [reduce ternary expressions](https://github.com/ethereum/solidity/issues/12930) to fewer opcodes and likely translate them to [`JUMPI`](https://www.evm.codes/#57). Therefore `TernaryLib` utils written in inline assembly have been added to replace ternary expressions as much as possible. `sortTokens` is also refactored to `TernaryLib` but inlined where appropriate.

In general for `x = y ? a : b`, when `y = true`
```Solidity
x = a = 0 ^ a = b ^ b ^ a = b ^ (b ^ a) * y
```
When `y = false`
```Solidity
x = b = b ^ 0 = b ^ (b ^ a) * y
```
Therefore `x = y ? a : b` is equivalent to `x = b ^ (b ^ a) * y` according to the properties of [`xor`](https://en.wikipedia.org/wiki/Exclusive_or).


contracts/modules/uniswap/v2/TernaryLib.sol
```diff
@@ -0,0 +1,42 @@
...
+    /// @notice Equivalent to the ternary operator: `condition ? a : b`
+    function ternary(bool condition, uint256 a, uint256 b) internal pure returns (uint256 res) {
+        assembly {
+            res := xor(b, mul(xor(a, b), condition))
+        }
+    }
+
+    /// @notice Sorts two uint256 in ascending order
+    /// @dev Equivalent to: `a < b ? (a, b) : (b, a)`
+    function sort2(uint256 a, uint256 b) internal pure returns (uint256, uint256) {
```
ewilz, line 15: Is this used anywhere?

contracts/modules/uniswap/v2/TernaryLib.sol
```diff
@@ -0,0 +1,42 @@
...
+        assembly {
+            let diff := mul(xor(tokenA, tokenB), lt(tokenB, tokenA))
+            token0 := xor(tokenA, diff)
+            token1 := xor(tokenB, diff)
+        }
+    }
+
+    /// @notice Swaps two uint256 if `condition` is true
+    /// @dev Equivalent to: `condition ? (b, a) : (a, b)`
+    function swapIf(bool condition, uint256 a, uint256 b) internal pure returns (uint256, uint256) {
```
ewilz, line 34: since this is a swap router, swap can easily get confused here. maybe `reverseOrderIf` / `reverseIf` / `switchIf` / `switchOrderIf`

contracts/modules/uniswap/v2/TernaryLib.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity >=0.5.0;
+
+/// @title Library for replacing ternary operator with efficient bitwise operations
+library TernaryLib {
```
ewilz, line 5: slick ternary ops in here!

contracts/modules/uniswap/v2/TernaryLib.sol
```diff
@@ -0,0 +1,42 @@
...
+    /// @notice Equivalent to the ternary operator: `condition ? a : b`
+    function ternary(bool condition, uint256 a, uint256 b) internal pure returns (uint256 res) {
```
ewilz, line 7: not used anywhere either?

contracts/modules/uniswap/v2/TernaryLib.sol
ewilz: One function is a ternary, while the other is not. I wonder if `SortLib` would be more readable and make more sense?

contracts/modules/uniswap/v2/UniswapV2Library.sol
```diff
@@ -52,15 +52,22 @@ library UniswapV2Library {
...
-        pair = address(
-            uint160(
-                uint256(
-                    keccak256(
-                        abi.encodePacked(hex'ff', factory, keccak256(abi.encodePacked(token0, token1)), initCodeHash)
-                    )
-                )
-            )
-        );
+        assembly ("memory-safe") {
```
ewilz, line 56: I think I like the gas savings here. Wonder if it's worth a comment:

```
/// accomplishes the following with with clean upper bits
///uint160(keccak256(abi.encodePacked(hex'ff', factory, keccak256(abi.encodePacked(token0, token1)), initCodeHash)))
```
to get a feel for what this is doing at a glance

contracts/modules/uniswap/v3/V3SwapRouter.sol
```diff
@@ -131,7 +141,12 @@ abstract contract V3SwapRouter is RouterImmutables, Permit2Payments, IUniswapV3S
         (int256 amount0Delta, int256 amount1Delta, bool zeroForOne) =
             _swap(-amountOut.toInt256(), recipient, path, payer, false);
 
-        uint256 amountOutReceived = zeroForOne ? uint256(-amount1Delta) : uint256(-amount0Delta);
+        uint256 amountOutReceived;
+        assembly {
+            // amountOutReceived = uint256(-(zeroForOne ? amount1Delta : amount0Delta))
+            // no need to check for underflow
+            amountOutReceived := sub(0, xor(amount0Delta, mul(xor(amount1Delta, amount0Delta), zeroForOne)))
```
ewilz, line 148: could these ternaries in this file get abstracted into the Library with all the xors??

contracts/modules/uniswap/v2/UniswapV2Library.sol
shuhuiluo: Added.

contracts/modules/uniswap/v3/V3SwapRouter.sol
shuhuiluo: Actually yes. I was afraid the optimizer wouldn't inline properly. But it seems the gas level remains the same as long as we uncheck the unary `-` on `int256`
```Solidity
unchecked {
    uint256 amountOutReceived = uint256(-zeroForOne.ternary(amount1Delta, amount0Delta));
    if (amountOutReceived != amountOut) revert V3InvalidAmountOut();
}
```

For
```Solidity
hasMultiplePools ? address(this) : recipient
```
using `hasMultiplePools.ternary(address(this), recipient)` in place also saves gas due to less stack shuffling.

For `zeroForOne := xor(isExactIn, lt(tokenOut, tokenIn))` though, using
```Solidity
zeroForOne = isExactIn.ternary(tokenIn < tokenOut, tokenOut < tokenIn);
```
wouldn't be cheaper. And writing `zeroForOne = isExactIn ^ (tokenOut < tokenIn)` isn't allowed since "Operator ^ not compatible with types bool and bool.".

Writing
```Solidity
sqrtPriceLimitX96 = zeroForOne.ternary(MIN_SQRT_RATIO + 1, MAX_SQRT_RATIO - 1);
```
instead of `xor` and literal hexes in assembly also wouldn't be cheaper. We could define the literal hexes as contract level constants though if that improves readability.

contracts/modules/uniswap/v2/TernaryLib.sol
shuhuiluo: It was more of an illustration. Now that overloaded versions of `ternary()` on `int256` and `address` are used in `V3SwapRouter`, should we leave this one here for completeness?

contracts/modules/uniswap/v2/TernaryLib.sol
shuhuiluo: No.

contracts/modules/uniswap/v2/TernaryLib.sol
shuhuiluo: I opt for `switchIf`.

contracts/modules/uniswap/v2/TernaryLib.sol
shuhuiluo: Only `sortTokens` is used for sorting. But the rest of the library is meant for replacement of the built-in ternary operator.

contracts/modules/uniswap/v2/UniswapV2Library.sol
ewilz: thanks,  I think that's pretty helpful for readability, one for the V3Lib would probably make sense too. (sorry, missed that)

contracts/modules/uniswap/v3/V3SwapRouter.sol
ewilz: cool thanks for the explanation. We definitely strive to strike a balance between readability and gas savings (or else we could just write low-level code for everything :)  

Even with our steepest gas savings tests, these ternaries that cannot be abstracted away would only save < 100 extra gas.  So I'd personally vote to just omit them. Can always get a third opinion..

Also some context, our v2 UniversalRouter has already been audited so we cannot include this, these gas savings would go into V3 of the router.

contracts/modules/uniswap/v2/UniswapV2Library.sol
shuhuiluo: Added.

contracts/modules/uniswap/v3/V3SwapRouter.sol
shuhuiluo: We can always abstract them away. It will just costs a few more opcodes. I'm okay with replacing `sqrtPriceLimitX96` done in assembly with literal hexes with `zeroForOne.ternary(MIN_SQRT_RATIO + 1, MAX_SQRT_RATIO - 1)`. But frankly speaking, is
```Solidity
zeroForOne = isExactIn ? tokenIn < tokenOut : tokenOut < tokenIn;
```
more understandable and cleaner than
```Solidity
assembly {
    zeroForOne := xor(isExactIn, lt(tokenOut, tokenIn))
}
```
with derivations?

PR #294, opened 2023-05-11
Decode everything in `calldata`

shuhuiluo:
Replace the rest of `abi.decode` with decoding to `calldata` using inline assembly. Reduce gas and bytecode size.


hensha256:
Thank you! Because this is a fork PR we cannot run CI, so I will merge this into a local branch and then set up a new PR where we can run CI before merging into main.

Issue #246, opened 2023-03-08
Url for `permit2` in `.gitmodules` causes sync issue


shuhuiluo:
The url of `permit2` in `.gitmodules` creates sync issue for me. It was `git@github.com:Uniswap/permit2.git`. But using it prevents GitHub Desktop or Foundry from installling dependencies automatically.

To replicate, simply do
```
git clone https://github.com/Uniswap/universal-router
cd .\universal-router\ 
forge install
```

The following is printed:
```
Cloning into '******/universal-router/lib/permit2'...
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

After changing the url to `https://github.com/Uniswap/permit2`, it syncs smoothly.

PR #296, opened 2023-05-19
change permit2 gh url

restlessronin:
@hensha256 I could not link the permit2 submodule until I made this change. I then had to `git submodule deinit lib/permit2` followed by `git submodule init`/


shuhuiluo:
Closes #246.

hensha256 (approved):

hensha256:
Thanks for doing this! CI will fail due to this being a fork PR - so our process is that we have to merge this into a local branch and make a new PR into main so we can allow CI to run in full. I'll sort that now!

PR #301, opened 2023-05-25
Pass tests with alternative FORK_URL

restlessronin:
@hensha256 I had trouble getting the tests working, apparently because my Infura account doesn't support archive nodes. I had to use the FORK_URL environment variable to set an alchemy url.

This worked for `forge test` but not `hardhat test`. I had to change a couple of places in the code where an Infura url was hardcoded as the fork url.

After these changes, all the tests work perfectly, both forge and hardhat.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,2 @@
 INFURA_API_KEY=
+FORK_URL=https://mainnet.infura.io/v3/${INFURA_API_KEY}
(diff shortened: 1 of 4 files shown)


restlessronin:
@hensha256 also added the FORK_URL to the sample .env file so new users know it's there

PR #314, opened 2023-08-24
feat: add new v2 deployments

marktoda:
V2 was deployed on more chains, this commit adds them to deploy
parameters for those chains


hensha256 (approved):
Missing polygon? Otherwise lgtm

PR #323, opened 2023-10-11
feat: add a Gnosis Chain deployment config

filoozom:
I'm not exactly sure what the two `InitCodeHash` variables are about. I see that `pairInitCodeHash` is not set for the testnets but that all mainnet ones are set to `0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f`, so I'm presuming that this is also the value we should use?

### DIFF
--- a/script/deployParameters/DeployGnosis.s.sol
+++ b/script/deployParameters/DeployGnosis.s.sol
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.15;
+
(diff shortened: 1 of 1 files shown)


filoozom:
We ended up deploying the contracts ourselves:
- Permit2: `0x80Ca57cAEFd5404be92710CE56Ef3b224dbF93E7`
- Unsupported protocol: `0xA174C4bb18b36B446C46d41E33a107B83686e92D`
- Universal router: `0xAe4b19d028C6FeF6Ba808454aDe47898B990b028`

PR #348, opened 2024-06-04
first pass

hensha256:
Remove NFT protocols to prepare for a new version that supports uniswap v2, v3 and v4


contracts/UniversalRouter.sol
```diff
@@ -79,6 +60,6 @@ contract UniversalRouter is IUniversalRouter, Dispatcher, RewardsCollector {
         return command & Commands.FLAG_ALLOW_REVERT == 0;
     }
 
-    /// @notice To receive ETH from WETH and NFT protocols
+    /// @notice To receive ETH from WETH protocols
```
ewilz, line 63: I think we can finally add this in now? 
```
require(msg.sender == WETH9, 'Not WETH9');
```

ewilz: that's been a huge thorn in our side

test/integration-tests/UniversalRouter.test.ts
```diff
@@ -149,225 +117,37 @@ describe('UniversalRouter', () => {
...
-        planner.addCommand(CommandType.SEAPORT_V1_4, [seaportValue, invalidSeaportCalldata], true)
-        const { commands, inputs } = planner
-
-        const miladyBalanceBefore = await miladyContract.balanceOf(alice.address)
-        await router['execute(bytes,bytes[],uint256)'](commands, inputs, DEADLINE, { value })
-        const miladyBalanceAfter = await miladyContract.balanceOf(alice.address)
-        expect(miladyBalanceAfter.sub(miladyBalanceBefore)).to.eq(numMiladys)
-      })
-    })
+    it('reverts if no commands are allowed to revert')
```
ewilz, line 150: unrun test cases?

contracts/UniversalRouter.sol
ewilz: we'll probably need to check that it's either WETH or v4 PoolManager

contracts/modules/Payments.sol
```diff
@@ -37,21 +35,6 @@ abstract contract Payments is PaymentsImmutables {
         }
     }
 
-    /// @notice Approves a protocol to spend ERC20s in the router
-    /// @param token The token to approve
-    /// @param spender Which protocol to approve
-    function approveERC20(ERC20 token, Spenders spender) internal {
```
ewilz, line 43: cool

ewilz (changes requested):
can we add that `receive() external payable {}` change???

test/integration-tests/UniversalRouter.test.ts
hensha256: yeeah need to completely rewrite them as they were all written for NFTs

PR #354, opened 2024-06-14
Reentrancy guard transient

hensha256:


test/integration-tests/gas-tests/__snapshots__/UniversalVSSwapRouter.gas.test.ts.snap
dianakocsis, line 1: 😍

contracts/libraries/Locker.sol
```diff
@@ -0,0 +1,21 @@
...
+
+/// @notice This is a temporary library that allows us to use transient storage (tstore/tload)
+/// TODO: This library can be deleted when we have the transient keyword support in solidity.
+library Locker {
+    // The slot holding the locker state, transiently. bytes32(uint256(keccak256("LockedBy")) - 1)
+    bytes32 constant LOCKED_BY_SLOT = 0x0aedd6bde10e3aa2adec092b02a3e3e805795516cda41f27aa145b8f300af87a;
+
+    function set(address locker) internal {
+        assembly {
+            tstore(LOCKED_BY_SLOT, locker)
```
dianakocsis, line 12: clear the upper bytes of the address?
`tstore(LOCKED_BY_SLOT, and(locker, 0xffffffffffffffffffffffffffffffffffffffff))`
can these also be made memory safe?

test/foundry-tests/Locker.t.sol
```diff
@@ -0,0 +1,24 @@
...
+
+        Locker.set(locker2);
+        assertEq(Locker.get(), locker2);
+
+        Locker.set(locker3);
+        assertEq(Locker.get(), locker3);
+
+        Locker.set(address(0));
+        assertEq(Locker.get(), address(0));
+    }
```
dianakocsis, line 23: maybe add another test to make sure the slot is correct

contracts/base/LockAndMsgSender.sol
```diff
@@ -2,19 +2,17 @@
 pragma solidity ^0.8.24;
 
 import {Constants} from '../libraries/Constants.sol';
+import {Locker} from '../libraries/Locker.sol';
 
 contract LockAndMsgSender {
```
dianakocsis, line 7: this contract seems a little silly lol

contracts/libraries/Locker.sol
hensha256: grrrrl ur amazing

contracts/base/LockAndMsgSender.sol
hensha256: It definitely is 😂 honestly i think the `map` function doesnt belong here... feels like the lock and the msg.sender logic should be separated

hensha256: ok i've moved `map` to the main dispatcher file because it doesnt feel like it belongs in the lock. And added a _msgSender function to make the lock msg.sender logic clearer

package.json
```diff
@@ -61,12 +61,12 @@
...
+    "compile": "hardhat compile && forge build",
+    "test:hardhat": "yarn compile && hardhat test",
     "test:gas": "UPDATE_SNAPSHOT=1 yarn test --grep gas",
-    "test:all": "UPDATE_SNAPSHOT=1 yarn test",
+    "test:all": "UPDATE_SNAPSHOT=1 yarn test:hardhat && forge test --isolate",
     "prettier:fix": "prettier --write '**/*.ts' && prettier --write '**/*.json'",
-    "lint:fix": "yarn prettier:fix && forge fmt",
-    "prettier": "prettier --check '**/*.ts' && forge fmt --check"
+    "lint": "yarn prettier:fix && forge fmt",
+    "lint:check": "prettier --check '**/*.ts' && forge fmt --check"
```
hensha256, line 70: renamed some commands here
`yarn test:all` wasnt running all the tests... only hardhat. So now `yarn test:hardhat` runs hardhat, and `yarn test:all` runs both
`yarn lint` should just fix everything locally both javascript and solidity then`yarn lint:check` can run the reverting commands which have to be run in CI

.github/workflows/lint.yml
```diff
@@ -40,4 +40,4 @@ jobs:
         id: build
 
       - name: Run linter
-        run: yarn run prettier
```
snreynolds, line 43: was this broken before?

contracts/base/LockAndMsgSender.sol
```diff
@@ -1,35 +1,27 @@
...
-        if (recipient == Constants.MSG_SENDER) {
-            return lockedBy;
-        } else if (recipient == Constants.ADDRESS_THIS) {
-            return address(this);
-        } else {
-            return recipient;
-        }
+    /// @notice Function to be used instead of msg.sender, as the contract performs self-reentrancy and at
+    /// times msg.sender == address(this). Instead _msgSender() returns the initiator of the lock
+    function _msgSender() internal view returns (address) {
```
snreynolds, line 24: I think this will have to override the _msgSender() in BaseActionsRouter..

snreynolds: when you add the v4 swap lib

contracts/base/LockAndMsgSender.sol
```diff
@@ -1,35 +1,27 @@
...
-import {Constants} from '../libraries/Constants.sol';
+import {Locker} from '../libraries/Locker.sol';
 
 contract LockAndMsgSender {
     error ContractLocked();
 
-    address internal constant NOT_LOCKED_FLAG = address(1);
-    address internal lockedBy = NOT_LOCKED_FLAG;
-
+    /// @notice Modifier enforcing a reentrancy lock that allows self-reentrancy
```
snreynolds, line 9: Can an external party that is not msg.sender trigger self reentrancy?

contracts/base/LockAndMsgSender.sol
```diff
@@ -1,35 +1,27 @@
...
     modifier isNotLocked() {
+        // The contract is allowed to reenter itself to perform EXECUTE_SUB_PLAN commands
```
snreynolds, line 11: tbh I'm not really sure what this means/what is the motivation for this comment

contracts/base/LockAndMsgSender.sol
hensha256: yeah it will, or maybe theyll both have to be overriden by 1 implementing function. Not 100% sure yet until i pull in v4 and see what compiles LOL

contracts/base/LockAndMsgSender.sol
hensha256: no theres only 1 entry point to the contract so once its locked only the locker has control

.github/workflows/lint.yml
hensha256: no just renamed it because its running both `forge fmt` and `prettier` so the name `lint` makes more sense than `prettier`

contracts/base/LockAndMsgSender.sol
hensha256: Just trying to explain the modifier because it performs no checks in the case where `msg.sender == address(this)`. I'll rewrite it a bit

snreynolds (approved):

contracts/libraries/Locker.sol
dianakocsis: so if its msg.sender, the upper bytes can never be dirty?

contracts/base/LockAndMsgSender.sol
dianakocsis: great!

PR #355, opened 2024-06-26
V3 liquidity commands

dianakocsis:


remappings.txt
```diff
@@ -1,5 +1,6 @@
 solmate/=lib/solmate/
 permit2/=lib/permit2/
 forge-std/=lib/forge-std/src/
-@openzeppelin/=node_modules/@openzeppelin
+@openzeppelin/contracts/token/ERC721/=lib/ERC721
+@openzeppelin/contracts-v4=node_modules/@openzeppelin/contracts-v4
```
hensha256, line 5: what exactly happened to lead to this setup with ERC721 logic inside a different folder?

script/deployParameters/DeploySepolia.s.sol
```diff
@@ -12,7 +12,8 @@ contract DeploySepolia is DeployUniversalRouter {
             v2Factory: 0xB7f907f7A9eBC822a80BD25E224be42Ce0A698A0,
             v3Factory: 0x0227628f3F023bb0B980b67D528571c95c6DaC1c,
             pairInitCodeHash: 0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f,
-            poolInitCodeHash: 0xe34f199b19b2b4f47f68442619d555527d244f78a3297ea89325f843f87b8b54
+            poolInitCodeHash: 0xe34f199b19b2b4f47f68442619d555527d244f78a3297ea89325f843f87b8b54,
+            v3NFTPositionManager: 0xC36442b4a4522E871399CD717aBDD847Ab11FE88
```
hensha256, line 16: did you check that the position manager is the same on all chains? i would expect it to be different on some

package.json
```diff
@@ -31,14 +31,15 @@
     "node": ">=14"
   },
   "dependencies": {
+    "@openzeppelin/contracts-v4": "npm:@openzeppelin/contracts@4.7.0",
```
hensha256, line 34: i htink i'd rather this just be normal like it used to be, and then create a new mapping for the erc721 to solve whatever the problem there was?

lib/ERC721/IERC721Enumerable.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+// taken from openzepplin@3.4.2 and updated the solidity version to fix remapping issues due to different location of the file
```
hensha256, line 2:
```suggestion
// taken from openzeppelin@3.4.2 and updated the solidity version to fix remapping issues due to different location of the file
```

same in the other copied files

contracts/modules/MigratorImmutables.sol
```diff
@@ -0,0 +1,17 @@
...
+
+import {INonfungiblePositionManager} from '@uniswap/v3-periphery/contracts/interfaces/INonfungiblePositionManager.sol';
+
+struct MigratorParameters {
+    address v3PositionManager;
+}
+
+contract MigratorImmutables {
+    /// @dev v3PositionManager address
+    INonfungiblePositionManager internal immutable V3POSITIONMANGER;
```
hensha256, line 12: nit: would like _ i think like `V3_POSITION_MANAGER`

test/foundry-tests/UniswapV2.t.sol
```diff
@@ -33,7 +36,8 @@ abstract contract UniswapV2Test is Test {
             v2Factory: address(FACTORY),
             v3Factory: address(0),
             pairInitCodeHash: bytes32(0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f),
-            poolInitCodeHash: bytes32(0)
+            poolInitCodeHash: bytes32(0),
+            v3NFTPositionManager: address(V3_POSITION_MANAGER)
```
hensha256, line 40: i feel like it can just be address(0) here given this file just tests uniswap v2

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1202,167 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+
+      const receipt = await tx.wait()
+
+      const transferEvent = receipt.events?.find((event) => event.event === 'IncreaseLiquidity')
+
+      if (transferEvent && transferEvent.args) {
+        tokenId = transferEvent.args.tokenId
+      }
+    })
+    it('erc721 permit', async () => {
```
hensha256, line 1248: have you run prettier? `yarn prettier:fix`

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1202,167 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+
+      position = await v3NFTPositionManager.positions(tokenId)
+      owed0 = position.tokensOwed0
+      owed1 = position.tokensOwed1
+
+      expect(owed0).to.eq(0)
+      expect(owed1).to.eq(0)
+    })
+
+    it('burn', async () => {
```
hensha256, line 1328: would love to see more test cases for each action. 
- can you burn multiple positions in the same tx?
- if you decrease your liquidity but dont collect, can a different user maliciously collect?
- what happens if someone who doesnt hold a position tries to burn someone elses position? like if A submits B's permit signature and tries to burn their token it shouldnt

any other attacks you can think of!

hensha256 (changes requested):

test/integration-tests/Uniswap.test.ts
dianakocsis: i have!

contracts/modules/Migrator.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.24;
+
+import {MigratorImmutables} from '../modules/MigratorImmutables.sol';
+import {INonfungiblePositionManager} from '@uniswap/v3-periphery/contracts/interfaces/INonfungiblePositionManager.sol';
+import {TokenAuthorized} from '../libraries/TokenAuthorized.sol';
+
+abstract contract Migrator is MigratorImmutables {
+    using TokenAuthorized for address;
```
hensha256, line 9: Thinking about naming for this, i think i'd like if it included the word `cache`
like
`if (spender.isAuthorizationCached(tokenId))` and `if (authorized.cacheAuthorization(tokenId))`
i feel like those read more easily than
`is (spender.getAuthorized(tokenId))` and `if (authorized) spender.setAuthorized(tokenId)`

Then library called... `TokenAuthorizationCache`? or something?

contracts/modules/Migrator.sol
```diff
@@ -0,0 +1,34 @@
...
+        if (spender.getAuthorized(tokenId)) {
+            return true;
+        } else {
+            address owner = V3_POSITION_MANGER.ownerOf(tokenId);
+            authorized = spender == owner || V3_POSITION_MANGER.getApproved(tokenId) == spender
+                || V3_POSITION_MANGER.isApprovedForAll(owner, spender);
+            if (authorized) {
+                spender.setAuthorized(tokenId);
+            }
+            return authorized;
```
hensha256, line 31: dont need this line! its already a named return variable :)
or you can remove the name from the return var, and initialize it above with `bool authorized = ...`

contracts/libraries/TokenAuthorized.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
+
+library TokenAuthorized {
```
hensha256, line 4: would love a comment above this! just like 1-2 lines explaining the purpose of the library

contracts/libraries/TokenAuthorized.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
```
hensha256, line 2: 0.8.24! needs to be for tstore

test/integration-tests/shared/getPermitNFTSignature.ts
```diff
@@ -0,0 +1,58 @@
+import { BigNumberish, constants, Signature } from 'ethers'
+import { splitSignature } from 'ethers/lib/utils'
+import { INonfungiblePositionManager } from '../../../typechain'
+import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers'
+
+export default async function getPermitNFTSignature(
```
hensha256, line 6: did you copy this function from somewhere else? just wondering

test/integration-tests/Uniswap.test.ts
```diff
@@ -27,17 +28,23 @@ import { RoutePlanner, CommandType } from './shared/planner'
...
 import { getPermitSignature, getPermitBatchSignature, PermitSingle } from './shared/protocolHelpers/permit2'
 import { encodePathExactInput, encodePathExactOutput } from './shared/swapRouter02Helpers'
+import getPermitNFTSignature from './shared/getPermitNFTSignature'
+import { FeeAmount } from '@uniswap/v3-sdk'
 const { ethers } = hre
 
 describe('Uniswap V2 and V3 Tests:', () => {
   let alice: SignerWithAddress
   let bob: SignerWithAddress
+  let eve: SignerWithAddress
```
hensha256, line 38: love

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1205,436 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+      // check that the USDC and WETH were spent
+      expect(usdcSpent).to.be.gte(0)
+      expect(wethSpent).to.be.gte(0)
+
+      const receipt = await tx.wait()
+
+      const transferEvent = receipt.events?.find((event) => event.event === 'IncreaseLiquidity')
+
+      if (transferEvent && transferEvent.args) {
+        tokenId = transferEvent.args.tokenId
```
hensha256, line 1247: shouldnt the setup fail if no event is found?

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1205,436 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+        const { v, r, s } = await getPermitNFTSignature(bob, v3NFTPositionManager, router.address, tokenId, MAX_UINT)
+
+        planner.addCommand(CommandType.ERC721_PERMIT, [router.address, tokenId, MAX_UINT, v, r, s])
+
+        // bob permits the router to authorize token
+        await executeRouter(planner)
+
+        expect((await v3NFTPositionManager.positions(tokenId)).operator).to.eq(router.address)
+      })
+      it('only owner of the token can permit another address', async () => {
```
hensha256, line 1261: please can you put empty lines between all these blocks?

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1205,436 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+      const receipt = await tx.wait()
+
+      const transferEvent = receipt.events?.find((event) => event.event === 'IncreaseLiquidity')
+
+      if (transferEvent && transferEvent.args) {
+        tokenId = transferEvent.args.tokenId
+      }
+    })
+    describe('erc721permit', () => {
+      it('erc721 permit', async () => {
```
hensha256, line 1251: more decriptive name please! even just "erc721 permit succeeds" is better :)

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1205,436 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+        const { v, r, s } = await getPermitNFTSignature(bob, v3NFTPositionManager, router.address, tokenId, MAX_UINT)
+
+        planner.addCommand(CommandType.ERC721_PERMIT, [router.address, tokenId, MAX_UINT, v, r, s])
+
+        // bob permits the router to authorize token
```
hensha256, line 1256:
```suggestion
        // bob permits the router to spend token
```

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1205,436 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+        await executeRouter(planner)
+
+        expect((await v3NFTPositionManager.positions(tokenId)).operator).to.eq(router.address)
+      })
+      it('only owner of the token can permit another address', async () => {
+        const { v, r, s } = await getPermitNFTSignature(eve, v3NFTPositionManager, router.address, tokenId, MAX_UINT)
+        planner.addCommand(CommandType.ERC721_PERMIT, [router.address, tokenId, MAX_UINT, v, r, s])
+
+        // bob is trying to permit the router using alice's signature
```
hensha256, line 1265: the signature is eve's not alice's currently

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1205,436 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+
+        planner.addCommand(CommandType.V3_POSM_CALL, [encodedCall])
+        await executeRouter(planner)
+
+        position = await v3NFTPositionManager.positions(tokenId)
+        liquidity = position.liquidity
+
+        expect(liquidity).to.eq(0)
+      })
+      it('cannot call decrease liquidity with improper signature', async () => {
```
hensha256, line 1298:
```suggestion
      it('cannot call decrease liquidity with improper function selector', async () => {
```

hensha256: when i read `signature` i thought it was talking about the permit signature

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1205,436 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+          '(uint256 tokenId,uint256 liquidity,uint256 amount0Min,uint256 amount1Min, uint256 deadline)'
+
+        const params = { tokenId: tokenId, liquidity: liquidity, amount0Min: 0, amount1Min: 0, deadline: 0 }
+
+        const abi = new ethers.utils.AbiCoder()
+        const encodedParams = abi.encode([DECREASE_LIQUIDITY_STRUCT], [params])
+        const functionSignature = ethers.utils
+          .id('decreaseLiquidity((uint256,uint128,uint256,uint256,uint256))')
+          .substring(0, 10)
+        const encodedCall = functionSignature + encodedParams.substring(2)
```
hensha256, line 1339: why do these parameters fail? would love a comment for what about this information is invalid to cause it to fail

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1205,436 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+
+        planner.addCommand(CommandType.V3_POSM_CALL, [encodedCall])
+
+        await expect(executeRouter(planner)).to.be.revertedWithCustomError(router, 'CallToV3PositionManagerFailed')
+      })
+      it('cannot call decrease liquidity if not approved', async () => {
```
hensha256, line 1345: would also love a test where no `ERC721_PERMIT` command is provided to show that fails too

hensha256 (changes requested):
Left loads of thoughts! But its looks great!

contracts/base/Dispatcher.sol
```diff
@@ -33,171 +37,217 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
...
-                        payerIsUser := calldataload(add(inputs.offset, 0x80))
-                    }
-                    address[] calldata path = inputs.toAddressArray(3);
-                    address payer = payerIsUser ? lockedBy : address(this);
-                    v2SwapExactOutput(map(recipient), amountOut, amountInMax, path, payer);
-                } else if (command == Commands.PERMIT2_PERMIT) {
-                    // equivalent: abi.decode(inputs, (IAllowanceTransfer.PermitSingle, bytes))
-                    IAllowanceTransfer.PermitSingle calldata permitSingle;
+
+                    erc721Permit(spender, tokenId, deadline, v, r, s);
```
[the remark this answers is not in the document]
dianakocsis: right i wasnt sure! i guess since its probably just going to be used to permit the router for migrating, we can just use address(this)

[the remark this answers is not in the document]
dianakocsis: right i wasnt sure! i guess since its probably just going to be used to permit the router for migrating, we can just use address(this)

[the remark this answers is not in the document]
dianakocsis: right i wasnt sure! i guess since its probably just going to be used to permit the router for migrating, we can just use address(this)

test/integration-tests/shared/getPermitNFTSignature.ts
dianakocsis: yes! i got it from here: https://github.com/Uniswap/v3-periphery/blob/697c2474757ea89fec12a4e6db16a574fe259610/test/shared/getPermitNFTSignature.ts

test/integration-tests/Uniswap.test.ts
dianakocsis: i needed to grab the tokenId from the event - is there another way?

remappings.txt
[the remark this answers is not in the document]
dianakocsis: i dont think that will work because v3-periphery INonfungiblePositionManager imports `'@openzeppelin/contracts/token/ERC721/IERC721Metadata.sol';` OZ v4 contracts have this file in a different folder so it can't find it. So @openzeppelin/contracts needs to be the name for the old version

yarn.lock
```diff
@@ -4,27 +4,27 @@
 
 "@ampproject/remapping@^2.1.0":
   version "2.2.0"
-  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
+  resolved "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz"
```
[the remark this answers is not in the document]
dianakocsis: i just regenerated the yarn.lock file. Is that better?

remappings.txt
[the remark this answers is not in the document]
hensha256: ohhhhh i see 🤔 thats annoying

contracts/base/Dispatcher.sol
```diff
@@ -33,171 +37,217 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
...
-                // 0x08 <= command < 0x10
+                // 0x10 <= command < 0x18
             } else {
-                if (command == Commands.V2_SWAP_EXACT_IN) {
-                    // equivalent: abi.decode(inputs, (address, uint256, uint256, bytes, bool))
-                    address recipient;
-                    uint256 amountIn;
-                    uint256 amountOutMin;
-                    bool payerIsUser;
+                if (command == Commands.ERC721_PERMIT) {
```
hensha256, line 212: why cant this command just be joined with the `V3_POSM_CALL` command? it just decodes the parameters, and then passes them straight through to the position manager right?

contracts/modules/Migrator.sol
```diff
@@ -0,0 +1,33 @@
...
+        return selector == INonfungiblePositionManager.decreaseLiquidity.selector
+            || selector == INonfungiblePositionManager.collect.selector
+            || selector == INonfungiblePositionManager.burn.selector;
+    }
+
+    function isAuthorizedForToken(address spender, uint256 tokenId) internal returns (bool authorized) {
+        if (spender.isAuthorizationCached(tokenId)) {
+            return true;
+        } else {
+            address owner = V3_POSITION_MANGER.ownerOf(tokenId);
```
hensha256, line 25: dont we need to check if the msg.sender is the owner too? Otherwise if i approve the UR for my tokens, you could call and burn my tokens right?

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1206,480 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+
+      let bobUSDCBalanceAfter = await usdcContract.balanceOf(bob.address)
+      let bobWETHBalanceAfter = await wethContract.balanceOf(bob.address)
+
+      let usdcSpent = bobUSDCBalanceBefore.sub(bobUSDCBalanceAfter)
+      let wethSpent = bobWETHBalanceBefore.sub(bobWETHBalanceAfter)
+
+      // check that the USDC and WETH were spent
+      expect(usdcSpent).to.be.gte(0)
+      expect(wethSpent).to.be.gte(0)
```
hensha256, line 1241: this doesnt actually check that any usdc was spent? because both can be 0.

I think you need to check that _at least_ one of them is strictly greater than 0. So just 
`expect(usdcSpent > 0 || wethSpent > 0)`

test/integration-tests/Uniswap.test.ts
```diff
@@ -27,17 +28,24 @@ import { RoutePlanner, CommandType } from './shared/planner'
...
+  let eve: SignerWithAddress
   let router: UniversalRouter
   let permit2: IPermit2
   let daiContract: Contract
   let wethContract: Contract
   let usdcContract: Contract
   let planner: RoutePlanner
+  let v3NFTPositionManager: INonfungiblePositionManager
+
+  let tokenId: BigNumber
```
hensha256, line 48: little bit unclear what this is in the grand scheme of _all_ UR tests. Maybe `positionIDv3`

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1206,480 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+    describe('erc721permit', () => {
+      it('erc721 permit succeeds', async () => {
+        const { v, r, s } = await getPermitNFTSignature(bob, v3NFTPositionManager, router.address, tokenId, MAX_UINT)
+
+        planner.addCommand(CommandType.ERC721_PERMIT, [ADDRESS_THIS, tokenId, MAX_UINT, v, r, s])
+
+        // bob permits the router to spend token
+        await executeRouter(planner)
+
+        expect((await v3NFTPositionManager.positions(tokenId)).operator).to.eq(router.address)
```
hensha256, line 1259: good to check before the call that this isnt true

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1206,480 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+
+      const receipt = await tx.wait()
+
+      const transferEvent = receipt.events?.find((event) => event.event === 'IncreaseLiquidity')
+
+      tokenId = transferEvent?.args?.tokenId
+    })
+
+    describe('erc721permit', () => {
+      it('erc721 permit succeeds', async () => {
```
hensha256, line 1251:
```suggestion
      it('erc721 permit succeeds passing in ADDRESS_THIS for router', async () => {
```

or something 🤷‍♀️

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1206,480 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+
+      it('erc721 permit also succeeds when passing in actual address', async () => {
+        const { v, r, s } = await getPermitNFTSignature(bob, v3NFTPositionManager, router.address, tokenId, MAX_UINT)
+
+        planner.addCommand(CommandType.ERC721_PERMIT, [router.address, tokenId, MAX_UINT, v, r, s])
+
+        // bob permits the router to spend token
+        await executeRouter(planner)
+
+        expect((await v3NFTPositionManager.positions(tokenId)).operator).to.eq(router.address)
```
hensha256, line 1270: again check before execution call that this isnt true

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1206,480 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+      })
+
+      it('only owner of the token can permit another address', async () => {
```
hensha256, line 1273: The reason this test fails is because eve isnt the owner of bob's token. So the signature over `tokenId` which is bob's doesnt work because she doesnt own it.
However I think if you add a test where Bob generates a signature, and eve submits bob's signature it will succeed in approving UR for bob's token. This is a correct behaviour but would be good to add a test to show it.

I think good to have both test cases - where 1 fails and the other passes and make it clear the difference between the 2 tests

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1206,480 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+
+        const encodedDecreaseCall = encodeDecreaseLiquidity(decreaseParams)
+
+        planner.addCommand(CommandType.V3_POSM_CALL, [encodedDecreaseCall])
+        await executeRouter(planner)
+
+        position = await v3NFTPositionManager.positions(tokenId)
+        liquidity = position.liquidity
+
+        expect(liquidity).to.eq(0)
```
hensha256, line 1307: would be great to check that `tokensOwed` in the position manager gas increased too to show that there are tokens to claim after the decrease

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1206,480 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+
+        const encodedDecreaseCall = encodeDecreaseLiquidity(decreaseParams)
+
+        planner.addCommand(CommandType.V3_POSM_CALL, [encodedDecreaseCall])
+
+        // bob is trying to use the token that is now owned by eve. he is not authorized to do so
+        await expect(executeRouter(planner)).to.be.revertedWithCustomError(router, 'NotAuthorizedForToken')
+      })
+
+      it('bob is permitted over the nft so he can call decrease even though he is not the owner', async () => {
```
hensha256, line 1408: nice

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1206,480 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+        const functionSignatureCollect = ethers.utils.id('collect((uint256,address,uint128,uint128))').substring(0, 10)
+        const encodedCollectCall = functionSignatureCollect + encodedCollectParams.substring(2)
+
+        planner.addCommand(CommandType.V3_POSM_CALL, [encodedDecreaseCall])
+        planner.addCommand(CommandType.V3_POSM_CALL, [encodedCollectCall])
+
+        await expect(executeRouter(planner)).to.be.revertedWithCustomError(router, 'CallToV3PositionManagerFailed')
+      })
+
+      it('cannot call collect if not approved', async () => {
```
hensha256, line 1603:
```suggestion
      it('cannot call collect if the router is not approved', async () => {
```

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1206,480 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+        let owed1 = position.tokensOwed1
+
+        expect(owed0).to.eq(0)
+        expect(owed1).to.eq(0)
+
+        let bobToken0BalanceAfter = await usdcContract.balanceOf(bob.address)
+        let bobToken1BalanceAfter = await wethContract.balanceOf(bob.address)
+
+        // bob is the recipient - he should have received the owed tokens
+        expect(bobToken0BalanceAfter).to.be.gt(bobToken0BalanceBefore)
```
hensha256, line 1482: you know exactly how much it should have increased by by looking up `tokensOwed` in the v3 position manager before calling collect

test/integration-tests/Uniswap.test.ts
```diff
@@ -1196,6 +1206,480 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+        planner.addCommand(CommandType.V3_POSM_CALL, [encodedDecreaseCall])
+        planner.addCommand(CommandType.V3_POSM_CALL, [encodedCollectCall])
+
+        await expect(executeRouter(planner)).to.be.revertedWithCustomError(router, 'CallToV3PositionManagerFailed')
+      })
+
+      it('cannot call collect if not approved', async () => {
+        // first we need to permit the router to spend the nft
+        let { v, r, s } = await getPermitNFTSignature(bob, v3NFTPositionManager, router.address, tokenId, MAX_UINT)
+        planner.addCommand(CommandType.ERC721_PERMIT, [ADDRESS_THIS, tokenId, MAX_UINT, v, r, s])
```
hensha256, line 1606: would love to see another test where
- bob approves the router, and calls decrease
- separate tx eve tries to call collect to take the tokens

hensha256 (commented):
Left some thoughts mostly around more tests but mostly looking great 👯

contracts/base/Dispatcher.sol
dianakocsis: hm but v3 position manager call always takes in tokenId as the first parameter, permit needs to take spender as the first parameter

test/integration-tests/Uniswap.test.ts
dianakocsis: The thing is that permit, decrease, and collect are all happening in one call, so i cant look up tokensOwed after decrease and before collect. ill add another test that breaks them up - that way i can check to make sure they are equal

contracts/libraries/TokenAuthorizationCache.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.24;
+
+/// @title a library to store spenders' authorization over v3 position token ids in transient storage
+/// @dev this library implements the equivalent of a mapping, as transient storage can only be accessed in assembly
+library TokenAuthorizationCache {
+    /// @notice calculates which storage slot a boolean should be stored in for a given spender and tokenId
+    function _computeSlot(address spender, uint256 tokenId) internal pure returns (bytes32 hashSlot) {
+        assembly ("memory-safe") {
+            mstore(0, spender)
```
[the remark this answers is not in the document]
dianakocsis: if it's always passed as msg.sender does that mean it wouldn't be dirty?

test/integration-tests/Uniswap.test.ts
dianakocsis: hm for some reason tokensOwed1 isnt equaling out

contracts/base/Dispatcher.sol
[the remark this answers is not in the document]
dianakocsis: if we remove spender, then tokenId would be the first parameter and it could all be in v3_posm_call
but then we would need to concatenate address(this) to inputs when we call the erc721permit function in v3_posm_call -> leads to more if else statements

contracts/libraries/TokenAuthorizationCache.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.24;
+
+/// @title a library to store spenders' authorization over v3 position token ids in transient storage
+/// @dev this library implements the equivalent of a mapping, as transient storage can only be accessed in assembly
+library TokenAuthorizationCache {
```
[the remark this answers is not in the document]
dianakocsis: i've been thinking something like:
1. I approve a hook contract for all of my v3 position nfts by calling setApprovalForAll()
2. I call the UR and execute erc721permit (to permit the UR), decreaseLiquidity, swapV4, collect
3. In the swapV4 call, it calls out to the hook contract which transfers my position nft to someone else, making me not the owner anymore
4. Collect still works since my authorization is cached and I receive the tokens owed even though I am not the owner anymore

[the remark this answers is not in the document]
hensha256: > Collect still works since my authorization is cached

But would collect actually still work? I feel like when the actual `collect` call is executed, postiion manager will check "is UR approved for the user". And the NEW user has to have also approved UR for this to happen.

But yes i think if this new user has also approved the UR then it would likely go through 🙃

contracts/base/Dispatcher.sol
hensha256: yeah good point 👍

hensha256: i do think theres no need to decode all the parameters though. It could just have data, check the selector is for permit, and then call without decoding everything?

dianakocsis: oh I see because we wouldn't have to check isAuthorizedForToken if the action is erc721permit

.gitmodules
```diff
@@ -7,3 +7,9 @@
 [submodule "lib/permit2"]
 	path = lib/permit2
 	url = git@github.com:Uniswap/permit2.git
+[submodule "lib/v3-periphery"]
+	path = lib/v3-periphery
+	url = https://github.com/uniswap/v3-periphery
+[submodule "lib/v4-periphery"]
```
hensha256, line 13: 👀 fine to merge it anyway though, i know its coming very soon haha

dianakocsis: lol yeah not used yet haha

contracts/base/Dispatcher.sol
```diff
@@ -33,171 +36,199 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
...
-                    address payer = payerIsUser ? lockedBy : address(this);
-                    v2SwapExactInput(map(recipient), amountIn, amountOutMin, path, payer);
-                } else if (command == Commands.V2_SWAP_EXACT_OUT) {
-                    // equivalent: abi.decode(inputs, (address, uint256, uint256, bytes, bool))
-                    address recipient;
-                    uint256 amountOut;
-                    uint256 amountInMax;
-                    bool payerIsUser;
+                if (command == Commands.ERC721_PERMIT) {
+                    (success, output) = address(V3_POSITION_MANAGER).call(inputs);
```
hensha256, line 212: this needs to check its the `permit` selector 👀 
currently you could call `decreaseLiq` on a token you arent meant to access

test/integration-tests/gas-tests/Uniswap.gas.test.ts
```diff
@@ -64,7 +75,9 @@ describe('Uniswap Gas Tests', () => {
     bob = (await ethers.getSigners())[1]
     daiContract = new ethers.Contract(DAI.address, TOKEN_ABI, bob)
     wethContract = new ethers.Contract(WETH.address, TOKEN_ABI, bob)
+    usdcContract = new ethers.Contract(USDC.address, TOKEN_ABI, bob)
     permit2 = PERMIT2.connect(bob) as IPermit2
+    v3NFTPositionManager = V3_NFT_POSITION_MANAGER.connect(bob) as INonfungiblePositionManager
```
hensha256, line 80: theres a few changes to this file that can be reverted now bc youve moved everything to new test file

contracts/base/Dispatcher.sol
hensha256: would be great to have a test that reverts passing another selector to permit too

dianakocsis: ✅

test/integration-tests/V3ToV4Migration.test.ts
```diff
@@ -0,0 +1,856 @@
...
+        }
+
+        const encodedDecreaseCall = encodeDecreaseLiquidity(decreaseParams)
+
+        planner.addCommand(CommandType.ERC721_PERMIT, [encodedDecreaseCall])
+
+        // trying to execute the permit commmand by calling decrease liquidity
+        await expect(
+          executeRouter(planner, bob, router, wethContract, daiContract, usdcContract)
+        ).to.be.revertedWithCustomError(router, 'InvalidAction')
```
hensha256, line 137: nice

test/integration-tests/V3ToV4Migration.test.ts
```diff
@@ -0,0 +1,856 @@
...
+
+        planner.addCommand(CommandType.ERC721_PERMIT, [encodedErc721PermitCall])
+
+        // eve generated a signature for bob's token - fails since eve is not the owner
+        await expect(
+          executeRouter(planner, bob, router, wethContract, daiContract, usdcContract)
+        ).to.be.revertedWithCustomError(router, 'ExecutionFailed')
+      })
+
+      it('other address can call permit on behalf of someone as long as owner of the token generated the signature properly', async () => {
```
hensha256, line 162: yessss nice

test/integration-tests/V3ToV4Migration.test.ts
```diff
@@ -0,0 +1,856 @@
...
+          executeRouter(planner, bob, router, wethContract, daiContract, usdcContract)
+        ).to.be.revertedWithCustomError(router, 'NotAuthorizedForToken')
+      })
+
+      it('bob is permitted over the nft so he can call decrease even though he is not the owner', async () => {
+        // transfer the token to eve
+        await v3NFTPositionManager.transferFrom(bob.address, eve.address, tokenIdv3)
+
+        // eve permits bob to spend the token
+        await v3NFTPositionManager.connect(eve).setApprovalForAll(bob.address, true)
```
hensha256, line 376: would be great to have another test just like this that just approves the 1 specific token (not approval for all)

dianakocsis: so the opposite? the router is approved for all and bob is permitted for the 1 specific token?

test/integration-tests/V3ToV4Migration.test.ts
```diff
@@ -0,0 +1,856 @@
...
+
+        // not approved on the collect call
+        await expect(
+          executeRouter(planner, eve, router, wethContract, daiContract, usdcContract)
+        ).to.be.revertedWithCustomError(router, 'NotAuthorizedForToken')
+      })
+    })
+
+    describe('burn liquidity', () => {
+      it('burn succeeds', async () => {
```
hensha256, line 806: and one more than burn fails if you arent an approved spender of the nft

test/integration-tests/V3ToV4Migration.test.ts
hensha256: still do PERMIT command so that doesnt need to change, just Bob is permitted for only 1 token

contracts/base/Dispatcher.sol
```diff
@@ -33,171 +37,207 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, Callbacks,
...
-                } else if (command == Commands.BALANCE_CHECK_ERC20) {
-                    // equivalent: abi.decode(inputs, (address, address, uint256))
-                    address owner;
-                    address token;
-                    uint256 minBalance;
-                    assembly {
-                        owner := calldataload(inputs.offset)
-                        token := calldataload(add(inputs.offset, 0x20))
-                        minBalance := calldataload(add(inputs.offset, 0x40))
+                    if (!isAuthorizedForToken(lockedBy, tokenId)) {
```
snreynolds, line 234: This is a little bit of an odd authorization scheme.

This function call checks `isAuthorizedForToken` which checks if the msg.sender is the owner, or the approved "operator" (the address that gets set in `Position` struct) or if the operator mapping in base ERC721 contract is set to true.

By using this authorization logic inside the migrator we essentially require 2 operators to already be set on the token. (One, this contract must be approved because its going to be doing the call on position manager, but now 2, the current msg.sender needs to be approved bc we read it again here.) OR we just force the msg.sender is always the owner) 

the other weird thing is that only one of those "approvals" is permittable (setting the operator address in the Position struct is the only way to do a signature based permit on the ERC721). This is not necessarily a problem with this design here except for the fact that maybe there is a case to not use the authorization scheme of the position manager contract, and instead have our own authorization mapping inside the router. Food for thought. FWIW I also think this is a shortcoming of the position manager

contracts/modules/MigratorImmutables.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.24;
+
+import {INonfungiblePositionManager} from '@uniswap/v3-periphery/contracts/interfaces/INonfungiblePositionManager.sol';
+
+struct MigratorParameters {
```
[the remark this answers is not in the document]
hensha256: because itll also include v4 position manager so this struct is just preemptively there i think.
Its following the pattern of naming of 
- [UniswapParameters](https://github.com/Uniswap/universal-router/blob/228f2d151a5fc99836d72ae00f81db92cdb44bd3/contracts/modules/uniswap/UniswapImmutables.sol#L4) plus UniswapImmutables
- [PaymentParameters](https://github.com/Uniswap/universal-router/blob/228f2d151a5fc99836d72ae00f81db92cdb44bd3/contracts/modules/PaymentsImmutables.sol#L7) plus PaymentImmutables
- [RouterParameters](https://github.com/Uniswap/universal-router/blob/6cf84c8c8889661c160606a064ef03a64547f20f/contracts/base/RouterImmutables.sol#L4) plus RouterImmutables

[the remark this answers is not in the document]
hensha256: Could change it but at this point it would be a bigger renaming overhaul of UR 😅

contracts/libraries/Commands.sol
```diff
@@ -38,6 +38,9 @@ library Commands {
     // The following constant defines one of the boundaries where the if blocks split commands
     uint256 constant SECOND_IF_BOUNDARY = 0x10;
```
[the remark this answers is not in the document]
hensha256: yeah we just thought it was easier to read in the dispatcher to have boundaries... but i see that its confusing to have it in the commands file

contracts/base/Dispatcher.sol
hensha256: we discussed on slack, going to document well

contracts/modules/MigratorImmutables.sol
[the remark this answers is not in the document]
dianakocsis: yeah i just followed the pattern from the others! and we'll have v4posm in there too

test/integration-tests/V3ToV4Migration.test.ts
```diff
@@ -404,6 +404,43 @@ describe('V3 to V4 Migration Tests:', () => {
...
+        let position = await v3NFTPositionManager.positions(tokenIdv3)
+        let liquidity = position.liquidity
+
+        const params = { tokenId: tokenIdv3, liquidity: liquidity, amount0Min: 0, amount1Min: 0, deadline: MAX_UINT }
+
+        const encodedDecreaseCall = encodeDecreaseLiquidity(params)
+
+        planner.addCommand(CommandType.V3_POSITION_MANAGER_CALL, [encodedDecreaseCall])
+
+        await executeRouter(planner, bob, router, wethContract, daiContract, usdcContract)
```
hensha256, line 442: does this need to be 2 calls to `executeRouter`? can it just be 1?

dianakocsis: oh yeah changed the one above it too

contracts/modules/MigratorImmutables.sol
```diff
@@ -0,0 +1,17 @@
...
+    address v3PositionManager;
+}
+
+contract MigratorImmutables {
+    /// @dev v3PositionManager address
```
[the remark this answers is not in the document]
dianakocsis: changed PaymentsImmutables too

contracts/libraries/Commands.sol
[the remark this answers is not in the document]
snreynolds: I guess you could make it a separate constant and just the only downside is we have two constants equal to the same value?

contracts/modules/MigratorImmutables.sol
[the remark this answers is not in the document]
snreynolds: sg

snreynolds (approved):

hensha256 (approved):

PR #359, opened 2024-07-23
V4 liquidity commands

dianakocsis:


test/integration-tests/shared/deployV4.ts
```diff
@@ -0,0 +1,18 @@
+import hre from 'hardhat'
+const { ethers } = hre
+
+import { PositionManager, PoolManager } from '../../../typechain'
+
+export async function deployV4PositionManager(): Promise<string> {
+  const positionManagerFactory = await ethers.getContractFactory('PositionManager')
+  const positionManager = (await positionManagerFactory.deploy(
+    await deployV4PoolManager()
```
hensha256, line 9: just to future proof... we are also going to have a parameter in the UR deploy of `v4PoolManager` (for the v4 swapping commands). So I think it might make sense to have this function take `poolManager` as a parameter instead. And then in the deployRouter function itll be able to have poolManager as a parameter too? if that makes sense?

like `deployRouter` will need to do
````
const poolManager = deployV4PoolManager()
const routerParameters = {
   v4PoolManager: poolManager
   v4PositionManager: await deployV4PositionManager(poolManager)
}
```

no need to actually add the `poolManager` parameter to the router yet, but just want to future-proof this function

hensha256 (approved):

PR #369, opened 2024-08-01
separate uniswap tests

hensha256:


dianakocsis (commented):
should this section be in v2/v3 file since they're not mixing v2 and v3 trades?
https://github.com/Uniswap/universal-router/blob/2d377cfbcfbc5246e2cd62f2a726e76f3b15614b/test/integration-tests/UniswapMixed.test.ts#L65

dianakocsis (approved):

PR #376, opened 2024-08-12
Add v4 to receive

hensha256:
Adding v4 pool manager to the receive function of UR so it can receive ETH from v4

### DIFF
--- a/contracts/base/Dispatcher.sol
+++ b/contracts/base/Dispatcher.sol
@@ -177,12 +177,12 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, V4SwapRout
                     } else if (command == Commands.WRAP_ETH) {
                         // equivalent: abi.decode(inputs, (address, uint256))
                         address recipient;
-                        uint256 amountMin;
+                        uint256 amount;
                         assembly {
                             recipient := calldataload(inputs.offset)
-                            amountMin := calldataload(add(inputs.offset, 0x20))
+                            amount := calldataload(add(inputs.offset, 0x20))
                         }
-                        Payments.wrapETH(map(recipient), amountMin);
+                        Payments.wrapETH(map(recipient), amount);
                     } else if (command == Commands.UNWRAP_WETH) {
                         // equivalent: abi.decode(inputs, (address, uint256))
                         address recipient;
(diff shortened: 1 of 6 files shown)


contracts/base/Dispatcher.sol
```diff
@@ -177,12 +177,12 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, V4SwapRout
                     } else if (command == Commands.WRAP_ETH) {
                         // equivalent: abi.decode(inputs, (address, uint256))
                         address recipient;
-                        uint256 amountMin;
+                        uint256 amount;
```
snreynolds, line 180: whats the name change motivation?

hensha256: the parameter is not a slippage check "minimum amount to wrap"... its just an amount to wrap. For unwrap and sweep it _is_ a slippage check so i think the naming must have been copied across. For wrapping its just an amount.

test/integration-tests/UniswapMixed.test.ts
```diff
@@ -35,6 +47,13 @@ describe('Uniswap V2 and V3 Tests:', () => {
   let wethContract: Contract
   let usdcContract: Contract
   let planner: RoutePlanner
+  let v4Planner: V4Planner
+  let v4PoolManager: PoolManager
+  let v4PositionManager: PositionManager
+
+  // current market ETH price at block
+  const USD_ETH_PRICE = 3820
+  const ONE_PERCENT = 38
```
snreynolds, line 56: nit: this is 1% OF the pinned usd_eth_price so I feel like you can just calculate it in line using `USD_ETH_PRICE` I just got confused reading this bc its not "one percent"literally

test/integration-tests/UniswapMixed.test.ts
```diff
@@ -516,6 +557,58 @@ describe('Uniswap V2 and V3 Tests:', () => {
       expect(ethBalanceAfter.sub(ethBalanceBefore)).to.eq(fullAmountOut.sub(gasSpent))
     })
 
+    it.only('ERC20 --> ERC20 split V4 and V4 different routes, with wrap, aggregate slippage', async () => {
```
snreynolds, line 560: remove .only?

test/integration-tests/UniswapMixed.test.ts
```diff
@@ -516,6 +557,58 @@ describe('Uniswap V2 and V3 Tests:', () => {
...
+      // DAI -> USDC -> WETH
+      // and DAI -> USDC -> ETH -> WETH
```
snreynolds, line 562: ooo ok this was a little confusing, I thought they were both routes INTO weth, but we are actually doing one into weth and one into eth and then a wrap, maybe just undo this last comment for the second router to weth

snreynolds (commented):
all nits

snreynolds (approved):
nice

PR #379, opened 2024-08-19
add different tests for increasing and add tests for forwarding eth

dianakocsis:


test/integration-tests/gas-tests/V3ToV4Migration.gas.test.ts
```diff
@@ -566,6 +736,22 @@ describe('V3 to V4 Migration Gas Tests', () => {
 
         planner.addCommand(CommandType.V4_POSITION_MANAGER_CALL, [calldata])
 
+        compact = (
```
hensha256, line 739: i think it needs a comment on each of these un-permit things too to show whats happening to call increase.

hensha256 (approved):
Do you think its a recommended practice to be performing permits like in these tests? Even if you unpermit afterwards?
Do you see security concerns with that?
I feel like because both contracts have reentrancy locks its probably ok? But makes me nervous

hensha256:
But
1. theyre tests and we dont plan to do this with our frontend, we'll just use `mint`
2. the solidity change looks good

dianakocsis:
There might be some weird cases because we would have to require that all commands require success. So if we
1. permit router
2. increase liquidity on a pool which calls out to a hook that transfers the token
3. unpermit router will revert because old owner doesn't own the token anymore
it would have to use the new owner's signature

PR #380, opened 2024-08-20
update periphery to main

dianakocsis:


hensha256:
can you pull in latest periphery again? and do snapshots?

PR #382, opened 2024-08-27
Oz L-09

dianakocsis:
## Related Issue
OZ L-09
https://defender.openzeppelin.com/v2/#/audit/43d4e491-c5b4-42ba-b0fc-7f5adc46a6b6/issues/L-09

## Description of changes
Use `abi.encodeCall` instead of `abi.encodeWithSelector` since `abi.encodeWithSelector` is not type safe and `abi.encodeWithSignature` is not typo safe

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "solidity.compileUsingRemoteVersion": "v0.8.26+commit.8a97fa7a"
+}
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "solidity.compileUsingRemoteVersion": "v0.8.26+commit.8a97fa7a"
```
hensha256, line 2: 👀 should this be in git ignore?

dianakocsis: omg woopsie daisy

PR #387, opened 2024-08-30
ABDK CVF106

dianakocsis:
## Related Issue
ABDK CVF-106

## Description of changes
`forge fmt` didn't like ABDK's suggested change of putting the comment on the same line as the else branch, but I cleaned up the Commands (back from Sara's suggestion) and added more comments

### DIFF
--- a/contracts/base/Dispatcher.sol
+++ b/contracts/base/Dispatcher.sol
@@ -133,7 +133,7 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, V4SwapRout
                         revert InvalidCommandType(command);
                     }
                 } else {
-                    // 0x08 <= command < 0x0f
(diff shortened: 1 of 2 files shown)


contracts/base/Dispatcher.sol
```diff
@@ -132,8 +132,8 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, V4SwapRout
                         // placeholder area for command 0x07
                         revert InvalidCommandType(command);
                     }
-                    // 0x08 <= command < 0x10
                 } else {
+                    // 0x08 <= command < 0x0f
```
hensha256, line 136: should this still be 0x10? why 0x0f?

dianakocsis: yess

hensha256 (approved):

PR #398, opened 2024-09-12
beta release

marktoda:
bump

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "publishConfig": {
     "access": "public"
   },
-  "version": "1.6.0",
+  "version": "1.6.1-beta.1",
   "keywords": [
     "uniswap",
     "router",


package.json
```diff
@@ -5,7 +5,7 @@
   "publishConfig": {
     "access": "public"
   },
-  "version": "1.6.0",
+  "version": "1.6.1-beta.1",
```
ewilz, line 8: not necessarily a patch, but I guess it doesn't matter how these beta versions are published?? will prob eventually be our ur 2.0

ewilz (approved):
with minor nit?? doesn't really matter..

package.json
marktoda: ohh UR 2.0

marktoda: im just gonna send for now we can fix up later lol

ewilz: yan with 2.0 we can finally match up our npm versions with our UniversalRouter versions!

PR #416, opened 2024-10-11
restrict increase, decrease, and burn and v4 initialize pool call

dianakocsis:


contracts/base/Dispatcher.sol
```diff
@@ -269,9 +271,40 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, V4SwapRout
...
                     (success, output) = address(V3_POSITION_MANAGER).call(inputs);
+                } else if (command == Commands.V4_INITIALIZE_POOL) {
+                    bytes4 selector;
+                    assembly {
+                        selector := calldataload(inputs.offset)
+                    }
+                    if (selector != PoolInitializer.initializePool.selector) {
+                        revert InvalidAction(selector);
+                    }
+                    (success, output) = address(V4_POSITION_MANAGER).call(inputs);
```
hensha256, line 282: What is the need for the call to go to the position manager? Wouldnt it be better to just call the pool manager directly? and instead of checking the selector just call `PoolManager.initialize(params)`? Or is there a reason to do so via the position manager?

contracts/base/Dispatcher.sol
```diff
@@ -269,9 +271,40 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, V4SwapRout
...
+                    bytes calldata slice = inputs[4:];
+                    (bytes calldata actions,) = slice.toBytes(0).decodeActionsRouterParams();
+
+                    uint256 numActions = actions.length;
+
+                    for (uint256 actionIndex = 0; actionIndex < numActions; actionIndex++) {
+                        uint256 action = uint8(actions[actionIndex]);
+
+                        if (
+                            action == Actions.INCREASE_LIQUIDITY || action == Actions.DECREASE_LIQUIDITY
```
hensha256, line 302: i think this should all be in a helper function thats called like `checkV4PositionManagerCall` inside `V3ToV4Migrator`, and then this contract can stay way simpler. And all this logic can just go into there. So here its just

```
} else if (command == Commands.V4_POSITION_CALL) {
    checkV4PositionManagerCall()
    (success, output) = address(V4_POSITION_MANAGER).call{value: address(this).balance}(inputs);
}
```

and i'd argue we shouldve done the same for v3 but that can be separate PR to refactor

contracts/base/Dispatcher.sol
```diff
@@ -269,9 +271,40 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, V4SwapRout
...
+                    bytes4 selector;
+                    assembly {
+                        selector := calldataload(inputs.offset)
+                    }
+                    if (selector != V4_POSITION_MANAGER.modifyLiquidities.selector) {
+                        revert InvalidAction(selector);
+                    }
+
+                    bytes calldata slice = inputs[4:];
+                    (bytes calldata actions,) = slice.toBytes(0).decodeActionsRouterParams();
```
hensha256, line 294: oh you never use the parameters its decoding... i feel like it would be possible instead (and more efficient) to try this...
`slice.toBytes(0).toBytes(0)`

contracts/modules/V3ToV4Migrator.sol
```diff
@@ -3,10 +3,18 @@ pragma solidity ^0.8.24;
 
 import {MigratorImmutables} from '../modules/MigratorImmutables.sol';
 import {INonfungiblePositionManager} from '@uniswap/v3-periphery/contracts/interfaces/INonfungiblePositionManager.sol';
+import {Actions} from '@uniswap/v4-periphery/src/libraries/Actions.sol';
+import {CalldataDecoder} from '@uniswap/v4-periphery/src/libraries/CalldataDecoder.sol';
 
 /// @title V3 to V4 Migrator
 /// @notice A contract that migrates liquidity from Uniswap V3 to V4
 abstract contract V3ToV4Migrator is MigratorImmutables {
+    using CalldataDecoder for bytes;
```
dianakocsis, line 12: this isn't necessary anymore right? since youre just doing .toBytes(0) instead of decodeActionsRouterParams()

contracts/base/Dispatcher.sol
```diff
@@ -24,8 +26,6 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, V4SwapRout
 
     error InvalidCommandType(uint256 commandType);
     error BalanceTooLow();
-    error InvalidAction(bytes4 action);
-    error NotAuthorizedForToken(uint256 tokenId);
```
hensha256, line 28: just moved to the v3 migrator module

contracts/modules/V3ToV4Migrator.sol
```diff
@@ -20,4 +28,33 @@ abstract contract V3ToV4Migrator is MigratorImmutables {
...
+            selector := calldataload(inputs.offset)
+        }
+        if (selector != V4_POSITION_MANAGER.modifyLiquidities.selector) {
+            revert InvalidAction(selector);
+        }
+
+        bytes calldata slice = inputs[4:];
+        // first bytes(0) extracts the unlockData parameter from modifyLiquidities
+        // second bytes(0) extracts the actions parameter from unlockData
+        bytes calldata actions = slice.toBytes(0).toBytes(0);
```
snreynolds, line 45: 🥵 tricky

snreynolds (commented):
I think seems good to me, but I noticed in the tests that we dont actually use initialize before the mint, we just have a gas test for it. Unless I'm missing a test file? Could we add a full e2e test encoding the initialize then the mint, then check the success?

contracts/modules/V3ToV4Migrator.sol
hensha256: added more comments

snreynolds (approved):

contracts/modules/V3ToV4Migrator.sol
```diff
@@ -20,4 +28,38 @@ abstract contract V3ToV4Migrator is MigratorImmutables {
...
+        // the second bytes(0) extracts the actions parameter from unlockData
+        bytes calldata actions = slice.toBytes(0).toBytes(0);
+
+        uint256 numActions = actions.length;
+
+        for (uint256 actionIndex = 0; actionIndex < numActions; actionIndex++) {
+            uint256 action = uint8(actions[actionIndex]);
+
+            if (
+                action == Actions.INCREASE_LIQUIDITY || action == Actions.DECREASE_LIQUIDITY
```
kmbarry1, line 58: IMO it would be clearer (and should save some bytecode size) to express this condition as:

`if (action != Actions.MINT_POSITION) {`

kmbarry1: Or is there an explicit desire to allow all the other actions, and just disallow these three? In that case I'd argue the error name is a bit misleading and be `ForbiddenAction()` or something like that.

kmbarry1: Thinking about it more, it's gotta be the second case, so ignore the original comment and consider an error naming tweak.

contracts/base/Dispatcher.sol
```diff
@@ -269,12 +269,21 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, V4SwapRout
                     }
 
                     (success, output) = address(V3_POSITION_MANAGER).call(inputs);
-                } else if (command == Commands.V4_POSITION_CALL) {
+                } else if (command == Commands.V4_INITIALIZE_POOL) {
+                    PoolKey calldata poolKey;
+                    uint160 sqrtPriceX96;
+                    assembly {
+                        poolKey := inputs.offset
+                        sqrtPriceX96 := calldataload(add(inputs.offset, 0xa0))
```
kmbarry1, line 277: It may be nice to leave a comment explaining the `0xa0` constant, it's a somewhat fragile pattern.

contracts/modules/V3ToV4Migrator.sol
hensha256: Yeah because we still want settling, taking, clearing as actions. We can definitely make the error better and leave a comment to make this clearer

HickupHH3: alternative check is `if (action < Actions.SETTLE && action != Actions.MINT_POSITION)`, not sure if it'll be cheaper

PR #417, opened 2024-10-15
allow permit2 to silently fail to avoid dos

dianakocsis:


test/integration-tests/UniswapMixed.test.ts
```diff
@@ -279,6 +279,41 @@ describe('Uniswap V2, V3, and V4 Tests:', () => {
...
+      const sig = await getPermitBatchSignature(BATCH_PERMIT, bob, permit2)
+
+      // transfer funds into DAI-USDC and DAI-USDT pairs to trade
+      // do not allow revert
+      planner.addCommand(CommandType.PERMIT2_PERMIT_BATCH, [BATCH_PERMIT, sig])
+
+      // allow revert
+      planner.addCommand(CommandType.PERMIT2_PERMIT_BATCH, [BATCH_PERMIT, sig], true)
+
+      await executeRouter(planner, bob, router, wethContract, daiContract, usdcContract)
```
hensha256, line 314: would love a check for what the new nonce is or something to show the tx succeeded.

test/integration-tests/UniswapV2.test.ts
```diff
@@ -77,6 +77,31 @@ describe('Uniswap V2 Tests:', () => {
...
+      }
+      const sig = await getPermitSignature(permit, bob, permit2)
+
+      // 1) permit the router to access funds, not allowing revert
+      planner.addCommand(CommandType.PERMIT2_PERMIT, [permit, sig])
+
+      // 2) permit the router to access funds again, allowing revert
+      planner.addCommand(CommandType.PERMIT2_PERMIT, [permit, sig], true)
+
+      await executeRouter(planner, bob, router, wethContract, daiContract, usdcContract)
```
hensha256, line 102: same here

contracts/base/Dispatcher.sol
```diff
@@ -186,7 +193,14 @@ abstract contract Dispatcher is Payments, V2SwapRouter, V3SwapRouter, V4SwapRout
                             permitSingle := inputs.offset
                         }
                         bytes calldata data = inputs.toBytes(6); // PermitSingle takes first 6 slots (0..5)
-                        PERMIT2.permit(msgSender(), permitSingle, data);
+                        (success, output) = address(PERMIT2).call(
+                            abi.encodeWithSignature(
+                                'permit(address,((address,uint160,uint48,uint48),address,uint256),bytes)',
```
hensha256, line 198: auditors: would love to know if theres a better practice as i know `encodeWithSignature` is considered error-prone. `permit` is an overloaded function so we cannot use `encodeCall`

kmbarry1: Nothing comes to mind, unfortunately.

HickupHH3: `encodeWithSelector` can't be used for the same reason. Could define a separate interface that uniquely defines the permit functions, but it'll complicate the issue as opposed to just inlining it.

PR #469, opened 2026-02-20
feat: add SwapProxy contract for permit2-less swap flow

marktoda:
---
## Summary
Introduces a stateless `SwapProxy` contract that enables a 2-transaction swap flow (approve + swap) without requiring Permit2 signed messages. This provides an alternative integration path for platforms that cannot or prefer not to support EIP-712 signed permits.
## Changes
- Add `SwapProxy.sol` contract that pulls ERC20 tokens from users directly into the Universal Router, then executes UR commands with `payerIsUser=false`
- Add `ISwapProxy.sol` interface with NatSpec documentation
- Add `DeploySwapProxy.s.sol` deployment script
- Add comprehensive Foundry tests covering:
  - Exact input swaps via proxy
  - Exact output swaps with sweep for unused tokens
  - Revert on insufficient approval
  - Revert on expired deadline
  - Token safety checks (no tokens stuck in proxy/router)
  - Gas comparison benchmark vs direct Permit2 flow
## How It Works
```
User -> approve SwapProxy (one-time)
User -> SwapProxy.execute() (each swap)
         |-> transferFrom(user -> UR, amount)
         |-> UR.execute(commands, inputs, deadline)
```
## Integration Notes
- All swap commands **must** use `payerIsUser=false`
- All recipient addresses **must** be the user's explicit address, NOT `MSG_SENDER` (which resolves to the proxy contract)
- For exact output swaps, include a `SWEEP` command to return unused input tokens
- This contract is for token-input actions; ETH-input actions should be sent directly to the UR


contracts/SwapProxy.sol
```diff
@@ -0,0 +1,44 @@
...
+/// @notice Enables 2-tx swap flow (approve + swap) without Permit2 signed messages
+/// @dev Transfers tokens from the user directly into the Universal Router, then
+///      executes UR commands with payerIsUser=false so the router uses its own balance.
+///      IMPORTANT: All swap commands MUST use payerIsUser=false.
+///      All recipient addresses MUST be the user's explicit address, NOT MSG_SENDER,
+///      because MSG_SENDER resolves to this proxy contract within the UR execution context.
+contract SwapProxy {
+    using SafeTransferLib for ERC20;
+
+    IUniversalRouter public immutable universalRouter;
```
hensha256: we could always add UR as a parameter to `execute`? So when we upgrade people dont have to reapprove a new proxy?

contracts/SwapProxy.sol
```diff
@@ -0,0 +1,44 @@
...
+        ERC20(token).safeTransferFrom(msg.sender, address(universalRouter), amount);
+        universalRouter.execute{value: msg.value}(commands, inputs, deadline);
+
+        uint256 ethBalance = address(this).balance;
+        if (ethBalance > 0) {
+            SafeTransferLib.safeTransferETH(msg.sender, ethBalance);
+        }
+    }
+
+    receive() external payable {}
```
hensha256: Why is this needed?

contracts/SwapProxy.sol
```diff
@@ -0,0 +1,44 @@
...
+    /// @param deadline The transaction deadline
+    function execute(address token, uint256 amount, bytes calldata commands, bytes[] calldata inputs, uint256 deadline)
+        external
+        payable
+    {
+        ERC20(token).safeTransferFrom(msg.sender, address(universalRouter), amount);
+        universalRouter.execute{value: msg.value}(commands, inputs, deadline);
+
+        uint256 ethBalance = address(this).balance;
+        if (ethBalance > 0) {
```
hensha256: why will anything ever be left?

contracts/SwapProxy.sol
marktoda: I was thinking exact output refunds, but I guess you can just set the recipient as the sweep :D

contracts/SwapProxy.sol
marktoda: same as above

contracts/SwapProxy.sol
marktoda: minimizing params, simpler to integrate but agree that is probably nicer

hensha256 (approved):

contracts/SwapProxy.sol
marktoda: payable is on execute function claudeson

hensha256 (approved):

PR #472, opened 2026-03-17
fix(v2): use actual balance change for FOT token slippage checks

hensha256:
---
## Summary
Fixes per-hop slippage checks in V2 swaps to correctly handle fee-on-transfer (FOT) tokens by measuring actual balance changes instead of using calculated amounts from reserves.
## Changes
- Measure recipient balance before/after swap when `minHopPriceX36` is enabled to get actual received amount
- Add `minHopPriceEnabled` cache variable to avoid repeated array length checks in the loop
- Skip balance measurement when `minHopPriceX36[i] == 0` (no slippage check for that hop)
- Fix snapshot test name from `poolManager bytecode size` to `UniversalRouter bytecode size`
- Update gas snapshots (~96-327 gas increase for V2 swaps due to additional balance checks)
## Notes
The previous implementation calculated expected output using `getAmountOut()` from reserves, but FOT tokens transfer less than the calculated amount due to their transfer fees. This caused the per-hop price check to use an inflated output value, potentially allowing swaps that should have been rejected or incorrectly rejecting valid swaps.

### DIFF
--- a/contracts/modules/uniswap/v2/V2SwapRouter.sol
+++ b/contracts/modules/uniswap/v2/V2SwapRouter.sol
@@ -24,7 +24,7 @@ abstract contract V2SwapRouter is UniswapImmutables, Permit2Payments {
             (address token0,) = UniswapV2Library.sortTokens(path[0], path[1]);
             uint256 finalPairIndex = path.length - 1;
             uint256 penultimatePairIndex = finalPairIndex - 1;
-            bool minHopPricesActive = minHopPriceX36.length != 0;
+            bool minHopPriceEnabled = minHopPriceX36.length != 0;
             for (uint256 i; i < finalPairIndex; i++) {
                 (address input, address output) = (path[i], path[i + 1]);
                 (uint256 reserve0, uint256 reserve1,) = IUniswapV2Pair(pair).getReserves();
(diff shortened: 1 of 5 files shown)


contracts/modules/uniswap/v2/V2SwapRouter.sol
```diff
@@ -44,7 +40,18 @@ abstract contract V2SwapRouter is UniswapImmutables, Permit2Payments {
                         UNISWAP_V2_FACTORY, UNISWAP_V2_PAIR_INIT_CODE_HASH, output, path[i + 2]
                     )
                     : (recipient, address(0));
-                IUniswapV2Pair(pair).swap(amount0Out, amount1Out, nextPair, new bytes(0));
+
+                // if minHopPrice is being used, we need to check output balance change
+                if (minHopPriceEnabled && minHopPriceX36[i] != 0) {
+                    uint256 recipientBalance = ERC20(output).balanceOf(nextPair);
+                    IUniswapV2Pair(pair).swap(amount0Out, amount1Out, nextPair, new bytes(0));
+                    amountOutput = ERC20(output).balanceOf(nextPair) - recipientBalance;
```
zhongeric, line 48: Prob out of scope but a weird FOT token might hit this often. A semi popular implementation for FOT tokens is to atomically swap the fee into ETH on every swap. If the pool which they use is the `nextPair` here, the token balance will change (in this example, be higher than expected), which would push price up and cause reverts with tight limits. 

Generally feel that this can be put into that category of weird token implementations but just flagging for discussion

snapshots/UniversalRouterTest.json
```diff
@@ -1,3 +1,3 @@
 {
-  "poolManager bytecode size": "24200"
+  "poolManager bytecode size": "24546"
```
zhongeric: why did this change?

contracts/modules/uniswap/v2/V2SwapRouter.sol
hensha256: yeah thanks for the flag. Feels worth flagging to the team choosing the per-hop prices given we can detect FOT ahead of time.

PR #489, opened 2026-06-25
feat: add UNWRAP_WETH_EXACT command and bound command input decoding

gretzke:
## Summary
- Add UNWRAP_WETH_EXACT (0x0f) taking (recipient, amount). It unwraps exactly amount WETH and reverts with InsufficientETH if the router holds less.
- The existing UNWRAP_WETH command is unchanged. It already covers full-balance unwraps with a minimum, so the new command has no minAmount and no CONTRACT_BALANCE sentinel.
- Remove the two unreachable InvalidCommandType fallbacks in the saturated 0x00-0x07 and 0x08-0x0f dispatcher blocks. The compiler does not eliminate them itself, removing them saves 19 bytes each.
- Add a 0x40 calldata length check for the new command, following #497.

Stacked on #497 (base is agent/bound-command-input-decoding) so the length check can build on its helper. Retarget to main after #497 merges.

## Testing
- forge test --isolate, 111 passed, includes the extended short-input case for the new command
- npx hardhat test test/integration-tests/UniversalRouter.test.ts, 14 passing, covers partial unwrap, external recipient, full-balance amount, zero amount, insufficient balance revert
- gas snapshots regenerated, new partial unwrap costs 49898 gas, router bytecode 22405 bytes with 2171 bytes of margin

<details>

## Summary
Adds a new `UNWRAP_WETH_EXACT` command (`0x0f`) for exact-amount WETH unwrapping, and hardens all command input decoding with bounds checks to prevent out-of-bounds calldata reads.
### `UNWRAP_WETH_EXACT` command
- **`Commands.sol`**: Define `UNWRAP_WETH_EXACT = 0x0f` in the previously unused command slot
- **`Payments.sol`**: Add `unwrapWETH9Exact(recipient, amount)` — unwraps exactly `amount` WETH, reverts with `InsufficientETH` if balance is insufficient, no-ops when `amount == 0`
- **`Dispatcher.sol`**: Decode and dispatch the new command with two parameters (`recipient`, `amount`)
- **`README.md`**: Update command table to include `UNWRAP_WETH_EXACT` and fill in other previously unlisted commands (`0x0e`–`0x14`)
- **`planner.ts`**: Add `UNWRAP_WETH_EXACT` enum value and ABI definition
### Input bounds checking
- **`Dispatcher.sol`**: Add `checkInputLength` guard to 15 static-layout commands; add `decodePermitBatch` with full bounds validation of the dynamic `PermitDetails` array; remove unreachable `revert InvalidCommandType` branches
- **`BytesLib.sol`**: Add `toLengthOffset(bytes, uint256, uint256)` overload that validates element size, head offset, relative length pointer, and uses `div(remaining, elementSize)` to prevent overflow in length checks
- **`V3ToV4Migrator.sol`**: Add minimum-length checks in `_checkV3PermitCall`, `_checkV3PositionManagerCall`, and `_checkV4PositionManagerCall`
- **`IUniversalRouter.sol`**: Correct NatSpec for `signedRouteContext` to reflect V4 hook verification requirements (check `msg.sender == PoolManager` and sender param is the router)
### Tests
- Add Foundry tests: `testRejectsShortStaticCommandInputs` (15 commands), `testRejectsShortDynamicCommandInputs` (5 commands), `testPermit2PermitBatchDecodesBoundedDetails`, `testPermit2PermitBatchRejectsTruncatedDetails`, `testExecuteSignedRejectsSmuggledTransferInput`, `testFuzzTransferRejectsShortInput`
- Add 5 Hardhat integration tests for `UNWRAP_WETH_EXACT`: partial unwrap, external recipient, full-balance unwrap, zero-amount no-op, revert on insufficient balance
- Add gas snapshot for `UNWRAP_WETH_EXACT` partial unwrap (~49,898 gas)
- Updated all gas snapshots (bytecode size 22177 → 22405; most existing paths see minor gas decreases from branch restructuring)
## Notes
- **Not a breaking change**: The existing `UNWRAP_WETH` (`0x0c`) command is untouched — legacy integrations continue to work without modification
- Unlike `UNWRAP_WETH` which unwraps the full contract balance and checks against a minimum, `UNWRAP_WETH_EXACT` unwraps a precise amount, leaving any remainder as WETH in the contract
- The bounds checks close a class of calldata-smuggling vectors where undersized inputs could cause commands to read past their intended calldata slice

</details>

### DIFF
--- a/contracts/base/Dispatcher.sol
+++ b/contracts/base/Dispatcher.sol
@@ -172,8 +172,6 @@ abstract contract Dispatcher is
                             portion := calldataload(add(inputs.offset, 0x40))
                         }
                         Payments.payPortionFullPrecision(token, map(recipient), portion);
-                    } else {
-                        revert InvalidCommandType(command);
                     }
                 } else {
                     // 0x08 <= command < 0x10
@@ -271,9 +269,16 @@ abstract contract Dispatcher is
                         }
                         success = (ERC20(token).balanceOf(owner) >= minBalance);
                         if (!success) output = abi.encodePacked(BalanceTooLow.selector);
-                    } else {
-                        // placeholder area for command 0x0f
-                        revert InvalidCommandType(command);
+                    } else if (command == Commands.UNWRAP_WETH_EXACT) {
+                        checkInputLength(inputs, 0x40);
+                        // equivalent: abi.decode(inputs, (address, uint256))
+                        address recipient;
+                        uint256 amount;
+                        assembly {
+                            recipient := calldataload(inputs.offset)
+                            amount := calldataload(add(inputs.offset, 0x20))
+                        }
+                        Payments.unwrapWETH9Exact(map(recipient), amount);
                     }
                 }
             } else {
--- a/README.md
+++ b/README.md
@@ -84,7 +84,21 @@ Each command is a `bytes1` containing the following 8 bits:
    ├──────┼───────────────────────────────┤
    │ 0x0d │  PERMIT2_TRANSFER_FROM_BATCH  │
    ├──────┼───────────────────────────────┤
-   │ 0x0e-│  -------                      │
+   │ 0x0e │  BALANCE_CHECK_ERC20          │
+   ├──────┼───────────────────────────────┤
+   │ 0x0f │  UNWRAP_WETH_EXACT            │
+   ├──────┼───────────────────────────────┤
+   │ 0x10 │  V4_SWAP                      │
+   ├──────┼───────────────────────────────┤
+   │ 0x11 │  V3_POSITION_MANAGER_PERMIT   │
+   ├──────┼───────────────────────────────┤
+   │ 0x12 │  V3_POSITION_MANAGER_CALL     │
+   ├──────┼───────────────────────────────┤
+   │ 0x13 │  V4_INITIALIZE_POOL           │
+   ├──────┼───────────────────────────────┤
+   │ 0x14 │  V4_POSITION_MANAGER_CALL     │
+   ├──────┼───────────────────────────────┤
+   │ 0x15-│  -------                      │
    │ 0x20 │                               │
    ├──────┼───────────────────────────────┤
    │ 0x21 │  EXECUTE_SUB_PLAN             │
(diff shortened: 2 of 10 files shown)


contracts/base/Dispatcher.sol
```diff
@@ -263,8 +263,18 @@ abstract contract Dispatcher is
                         }
                         success = (ERC20(token).balanceOf(owner) >= minBalance);
                         if (!success) output = abi.encodePacked(BalanceTooLow.selector);
+                    } else if (command == Commands.UNWRAP_WETH_AMOUNT) {
```
dianakocsis: what if we made this UNWRAP_WETH_EXACT with only recipient, amount

dianakocsis: minAmount doesn't seem needed here anymore

contracts/base/Dispatcher.sol
```diff
@@ -263,8 +263,18 @@ abstract contract Dispatcher is
...
+                        address recipient;
+                        uint256 amount;
+                        uint256 minAmount;
+                        assembly {
+                            recipient := calldataload(inputs.offset)
+                            amount := calldataload(add(inputs.offset, 0x20))
+                            minAmount := calldataload(add(inputs.offset, 0x40))
+                        }
+                        Payments.unwrapWETH9(map(recipient), amount, minAmount);
                     } else {
```
dianakocsis: this can never be hit now

contracts/base/Dispatcher.sol
dianakocsis: also thoughts on calling this UNWRAP_WETH_EXACT instead?

contracts/base/Dispatcher.sol
```diff
@@ -263,8 +263,18 @@ abstract contract Dispatcher is
                         }
                         success = (ERC20(token).balanceOf(owner) >= minBalance);
                         if (!success) output = abi.encodePacked(BalanceTooLow.selector);
+                    } else if (command == Commands.UNWRAP_WETH_AMOUNT) {
+                        // equivalent: abi.decode(inputs, (address, uint256, uint256))
```
dianakocsis: also need to check length now like here: https://github.com/Uniswap/universal-router/pull/497

dianakocsis (approved):

PR #497, opened 2026-07-30
fix: bound router command input decoding

gretzke:
## Summary

This change makes Universal Router command decoding respect each input's declared calldata boundary.

- Add minimum ABI head checks before raw dispatcher reads.
- Validate Permit2 batch struct and array bounds before assigning calldata pointers.
- Harden shared dynamic-array decoding against out-of-bounds offsets, truncated elements, and arithmetic overflow.
- Add selector and token ID bounds checks to position manager call validation.
- Correct the V4 hook guidance to distinguish the PoolManager caller from the hook sender parameter.
- Add regression coverage for zero-length signed input smuggling, short static and dynamic command inputs, and truncated Permit2 batches.

## Root cause

`executeSigned` hashes each input using its declared `bytes` length. Several dispatcher paths used `calldataload` beyond that declared length. Noncanonical outer ABI encoding could therefore leave executable command parameters after an empty or truncated signed input.

The fix is applied at the command decoder boundary so the same invariant holds for signed and regular execution.

## Validation

- `forge build --sizes`, UniversalRouter runtime size is 22,093 bytes with 2,483 bytes of margin.
- `FORK_URL=<configured> forge test --isolate -vv`, 111 tests passed.
- Selected Permit2 Hardhat integration tests, 3 tests passed.
- Scoped `forge fmt --check` and `forge lint` passed for all changed Solidity files.

The repository-wide formatting check remains blocked by the existing formatting of `test/foundry-tests/permissionedPools/PermissionedV4.t.sol`, which is unchanged by this PR.

<details>

## Summary
Makes Universal Router command decoding respect each input's declared calldata boundary, preventing noncanonical outer ABI encoding from smuggling executable command parameters after an empty or truncated signed input. Also adds `UNWRAP_WETH_EXACT` (`0x0f`) for exact-amount WETH unwrapping.
## Root Cause
`executeSigned` hashes each input using its declared `bytes` length. Several dispatcher paths used `calldataload` beyond that declared length. Noncanonical outer ABI encoding could therefore leave executable command parameters after an empty or truncated signed input. The fix is applied at the command decoder boundary so the same invariant holds for signed and regular execution.
## Changes
- **Dispatcher**: Add `checkInputLength` minimum ABI head checks before raw `calldataload` reads for 15 static-head commands (V3/V2 swaps, PERMIT2_TRANSFER_FROM, SWEEP, TRANSFER, PAY_PORTION, WRAP_ETH, UNWRAP_WETH, BALANCE_CHECK_ERC20, V4_INITIALIZE_POOL, PERMIT2_PERMIT)
- **Dispatcher**: Replace inline Permit2 batch pointer assignment with `decodePermitBatch()` that validates struct offset, details array offset, and element count against input bounds
- **Dispatcher**: Remove two unreachable `revert InvalidCommandType` fallbacks in the saturated 0x00-0x07 and 0x08-0x0f dispatcher blocks
- **Dispatcher**: Decode SWEEP `amountMinimum` at full `uint256` width (was `uint160`)
- **Dispatcher**: Add `UNWRAP_WETH_EXACT` (`0x0f`) taking `(recipient, amount)`, which unwraps an exact amount of WETH and reverts with `InsufficientETH` if the balance is insufficient
- **BytesLib**: Add `toLengthOffset(bytes, uint256, uint256)` overload accepting an element size parameter; harden both variants against out-of-bounds offsets, truncated elements, and arithmetic overflow (division before multiplication)
- **V3ToV4Migrator**: Add minimum-length checks in `_checkV3PermitCall`, `_checkV3PositionManagerCall`, and `_checkV4PositionManagerCall` before `calldataload`
- **Payments**: Add `unwrapWETH9Exact(recipient, amount)` for exact-amount WETH unwrapping
- **IUniversalRouter**: Correct V4 hook NatSpec to distinguish the PoolManager caller (`msg.sender`) from the hook sender parameter
- **README**: Update command table to include `UNWRAP_WETH_EXACT` and fill in previously unlisted commands (`0x0e`-`0x14`)
- **planner.ts**: Add `UNWRAP_WETH_EXACT` enum value and ABI definition
## Tests
- Foundry: `testRejectsShortStaticCommandInputs` (15 commands), `testRejectsShortDynamicCommandInputs` (5 commands), `testPermit2PermitBatchDecodesBoundedDetails`, `testPermit2PermitBatchRejectsTruncatedDetails`, `testExecuteSignedRejectsSmuggledTransferInput`, `testFuzzTransferRejectsShortInput`, `testSweepTokenEnforcesFullWidthMinimum`, `testSweepTokenAtUint160MaxMinimum`
- Hardhat: 5 integration tests for `UNWRAP_WETH_EXACT` (partial unwrap, external recipient, full-balance, zero-amount no-op, insufficient balance revert)
- Gas snapshot for `UNWRAP_WETH_EXACT` partial unwrap added
- All gas snapshots regenerated
## Validation
- `forge build --sizes`: UniversalRouter runtime size is 22,093 bytes with 2,483 bytes of margin
- `forge test --isolate -vv`: 111 tests passed
- Selected Permit2 Hardhat integration tests: 3 tests passed
- `forge fmt --check` and `forge lint` passed for all changed Solidity files

</details>

### DIFF
--- a/contracts/base/Dispatcher.sol
+++ b/contracts/base/Dispatcher.sol
@@ -172,8 +172,6 @@ abstract contract Dispatcher is
                             portion := calldataload(add(inputs.offset, 0x40))
                         }
                         Payments.payPortionFullPrecision(token, map(recipient), portion);
-                    } else {
-                        revert InvalidCommandType(command);
                     }
                 } else {
                     // 0x08 <= command < 0x10
@@ -271,9 +269,16 @@ abstract contract Dispatcher is
                         }
                         success = (ERC20(token).balanceOf(owner) >= minBalance);
                         if (!success) output = abi.encodePacked(BalanceTooLow.selector);
-                    } else {
-                        // placeholder area for command 0x0f
-                        revert InvalidCommandType(command);
+                    } else if (command == Commands.UNWRAP_WETH_EXACT) {
+                        checkInputLength(inputs, 0x40);
+                        // equivalent: abi.decode(inputs, (address, uint256))
+                        address recipient;
+                        uint256 amount;
+                        assembly {
+                            recipient := calldataload(inputs.offset)
+                            amount := calldataload(add(inputs.offset, 0x20))
+                        }
+                        Payments.unwrapWETH9Exact(map(recipient), amount);
                     }
                 }
             } else {
--- a/contracts/base/Dispatcher.sol
+++ b/contracts/base/Dispatcher.sol
@@ -131,7 +131,7 @@ abstract contract Dispatcher is
                         // equivalent:  abi.decode(inputs, (address, address, uint256))
                         address token;
                         address recipient;
-                        uint160 amountMin;
+                        uint256 amountMin;
                         assembly {
                             token := calldataload(inputs.offset)
                             recipient := calldataload(add(inputs.offset, 0x20))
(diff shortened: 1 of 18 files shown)


contracts/base/Dispatcher.sol
```diff
@@ -317,6 +326,54 @@ abstract contract Dispatcher is
...
+        if (input.length < minimumLength) revert CalldataDecoder.SliceOutOfBounds();
+    }
+
+    /// @dev Decodes a Permit2 batch after validating its dynamic details array against the input bounds.
+    function decodePermitBatch(bytes calldata input)
+        private
+        pure
+        returns (IAllowanceTransfer.PermitBatch calldata permitBatch)
+    {
+        checkInputLength(input, 0x40);
```
dianakocsis: Could make this 0x60

contracts/base/Dispatcher.sol
dianakocsis: should be uint256

dianakocsis (approved):

PR #502, opened 2026-08-25
fix: gate nested V4_SWAP behind an explicit opt-in entrypoint

dianakocsis:
## What

#491 let a `V4_SWAP` run inside a PoolManager unlock opened by another contract whenever `poolManager.isUnlocked()` was true — inferring consent from chain state rather than from the caller.

All V4 deltas accrue under the router's address for the lifetime of that unlock, while the router's own locker clears between external calls. So one call could leave unpaid debt that a later call's `SETTLE_ALL` or `OPEN_DELTA` settlement paid out of a different caller's funds. That is audit finding **M-01**.

Consent now lives on the entrypoint:

```solidity
function executeNested(bytes calldata commands, bytes[] calldata inputs, uint256 deadline) external payable;
```

It sets a transient flag for the duration of the call; the nested branch in `Dispatcher` requires it. Plain `execute` / `executeSigned` now revert `NestedExecutionNotPermitted` instead of silently nesting.

## Why an entrypoint, and not a flag in the payload

Two independent reasons:

1. `EXECUTE_SIGNED_TYPEHASH` commits to `keccak256(commands)`, so an integrator carrying a signed route cannot alter its bytes to opt in.
2. More decisively, and this applies to unsigned routes too: with a payload flag the **route author** decides whether to nest. A composer that forwards a caller-supplied route would get nesting enabled without consenting — and the composer is exactly who M-01 puts at risk. Consent has to sit with the composer, which only an entrypoint provides.

## Accepted, not enforced

An opted-in caller owns the delta hygiene the router can no longer guarantee for it. A contract exposing a permissionless function that reaches `executeNested` can have its own capital drained by a route that leaves an `OPEN_DELTA`. That is a deliberate decision — the router does not re-assert a per-plan delta checkpoint. The integrator pattern is a function that does not open its own unlock checking for one and reverting.

## Known limitation: signed routes cannot nest

There is no `executeSignedNested` in this version, which is a **regression against `main`**, where `isUnlocked()` auto-detection let signed routes nest.

It costs 252 bytes against 473 remaining, and the RESOLVE proposal needs the same headroom. It matters for a composer whose route touches a hook that authorizes on the signer rather than on `msgSender()` — for example a permissioned pool wanting to check the human behind a composer, since `PermissionedV4Router` currently checks `msgSender()` and therefore sees the composer. Restoring it needs a new router deployment.

## Also carried: the calldata fix

The four v4 swap parameter decoders followed struct and member offsets bounded only by `calldatasize()`, not by `params.length`. The ordinary path was unaffected because `poolManager.unlock()` re-encodes on the way to `unlockCallback`; the nested branch forwarded a slice of the original transaction calldata, letting the decoders read bytes past `inputs[i].length` that no signature covers.

This branch first mitigated that with a self-call that re-encoded the input. #505 then bumped `lib/v4-periphery` to decoders that use `abi.decode`, which bound every struct and member offset against the input length — so #513 removed the self-call as redundant. The decoders now end the read at the input boundary on both paths.

Coverage is `V4NestedUnlockCalldataSmuggling.t.sol` (struct-offset redirect, both paths) and `V4NestedMemberSmuggle.t.sol` (member-offset redirect, the variant a struct-offset-only fix misses). Both blocked.

## Bytecode

| | Runtime | Margin |
|---|---|---|
| Before | 23,705 | 871 |
| After (incl. #505 bump and #513 removal) | 24,353 | **223** |

`optimizer_runs` is already pinned to `1` for `UniversalRouter.sol`, so there is no optimizer headroom left, and EIP-7907 is not in Glamsterdam's SFI list.

## Testing

`V4SwapWithinUnlock.t.sol` is 18 passing. The mocks are real composer shapes: `LockHolder` opens its own unlock and calls the router from its callback, `OuterLockOpener` opens the unlock and has a different contract call in, `CreditOuterLockOpener` creates its own credit first and checks it afterward.

New coverage:

- `test_v4Swap_withinExistingUnlock_plainExecuteReverts` — the M-01 victim shape, now blocked.
- `test_executeNested_optInDoesNotLeakToLaterCall` — the flag does not survive the call that set it.
- `test_executeNested_optInPersistsIntoSubPlan` — it does survive `EXECUTE_SUB_PLAN`'s `address(this).call` self-re-entry.

The pre-existing isolation assertions were re-pointed at `executeNested` rather than deleted, so a nested route still provably cannot drain the outer owner's credit or capture their in-flight synced tokens.

Gas: nested V4 swaps cost ~69 more gas; the Hardhat gas snapshots are unchanged, as they do not exercise the nested path.

- `forge test` — 139 passed, 0 failed
- `yarn test:hardhat` — 190 passing, 2 pending, 0 failed
- `yarn lint:check` — clean

<details>

## What
#491 let a `V4_SWAP` run inside a PoolManager unlock opened by another contract whenever `poolManager.isUnlocked()` was true — inferring consent from chain state rather than from the caller.
All V4 deltas accrue under the router's address for the lifetime of that unlock, while the router's own locker clears between external calls. So one call could leave unpaid debt that a later call's `SETTLE_ALL` or `OPEN_DELTA` settlement paid out of a different caller's funds. That is audit finding **M-01**.
Consent now lives on the entrypoint:
```solidity
function executeNested(bytes calldata commands, bytes[] calldata inputs, uint256 deadline) external payable;
```
It sets a transient flag for the duration of the call; the nested branch in `Dispatcher` requires it. Plain `execute` / `executeSigned` now revert `NestedExecutionNotPermitted` instead of silently nesting.
## Why an entrypoint, and not a flag in the payload
Two independent reasons:
1. `EXECUTE_SIGNED_TYPEHASH` commits to `keccak256(commands)`, so an integrator carrying a signed route cannot alter its bytes to opt in.
2. More decisively, and this applies to unsigned routes too: with a payload flag the **route author** decides whether to nest. A composer that forwards a caller-supplied route would get nesting enabled without consenting — and the composer is exactly who M-01 puts at risk. Consent has to sit with the composer, which only an entrypoint provides.
## Changes
- **`contracts/UniversalRouter.sol`**: Add `executeNested(commands, inputs, deadline)` entrypoint that sets a transient opt-in flag via `NestedUnlock.set(true)`, delegates to `execute`, then clears the flag
- **`contracts/base/Dispatcher.sol`**: Gate the nested `V4_SWAP` branch behind `NestedUnlock.isPermitted()`, reverting with `NestedExecutionNotPermitted` when the flag is not set; add comments explaining the calldata safety invariant provided by `abi.decode` decoders
- **`contracts/interfaces/IUniversalRouter.sol`**: Add `executeNested` to the interface with NatSpec documenting the delta-sharing risk model
- **`contracts/libraries/NestedUnlock.sol`**: New library using `tstore`/`tload` on a derived transient slot to record the opt-in state
- **`lib/v4-periphery`**: Bump submodule from `07336f21` to `0cc6e164` (`abi.decode` swap param decoders)
- **`snapshots/UniversalRouterTest.json`**: Bytecode size updated 23,705 → 24,353
## Accepted, not enforced
An opted-in caller owns the delta hygiene the router can no longer guarantee for it. A contract exposing a permissionless function that reaches `executeNested` can have its own capital drained by a route that leaves an `OPEN_DELTA`. The router does not re-assert a per-plan delta checkpoint. Credit is equally unsafe: positive delta left under the router by one call is takeable by any contract that gains control before the next call.
## Known limitation: signed routes cannot nest
There is no `executeSignedNested` in this version, which is a **regression against `main`**, where `isUnlocked()` auto-detection let signed routes nest.
## Also carried: the calldata fix
The four v4 swap parameter decoders followed struct and member offsets bounded only by `calldatasize()`, not by `params.length`. The ordinary path was unaffected because `poolManager.unlock()` re-encodes on the way to `unlockCallback`; the nested branch forwarded a slice of the original transaction calldata, letting the decoders read bytes past `inputs[i].length` that no signature covers.
This branch first mitigated that with a self-call that re-encoded the input. #505 then bumped `lib/v4-periphery` to decoders that use `abi.decode`, which bound every struct and member offset against the input length — so #513 removed the self-call as redundant. The decoders now end the read at the input boundary on both paths.
Coverage is `V4NestedUnlockCalldataSmuggling.t.sol` (struct-offset redirect, both paths) and `V4NestedMemberSmuggle.t.sol` (member-offset redirect, the variant a struct-offset-only fix misses). Both blocked.
## Bytecode
| | Runtime | Free under 24,576 |
|---|---|---|
| Before | 23,705 | 871 |
| After | 24,353 | **223** |
+648 total (includes both the opt-in gate and the v4-periphery submodule bump to `abi.decode` decoders).
## Testing
`V4SwapWithinUnlock.t.sol` extended with:
- `test_v4Swap_withinExistingUnlock_plainExecuteReverts` — the M-01 victim shape, now blocked
- `test_executeNested_optInDoesNotLeakToLaterCall` — the flag does not survive the call that set it
- `test_executeNested_optInPersistsIntoSubPlan` — it does survive `EXECUTE_SUB_PLAN`'s `address(this).call` self-re-entry
New test files:
- `V4NestedUnlockCalldataSmuggling.t.sol` — regression coverage for struct-offset calldata smuggling on both nested and non-nested paths
- `V4NestedMemberSmuggle.t.sol` — member-offset smuggling variant, confirms `abi.decode` decoders block reads past `inputs[i].length`
Pre-existing isolation assertions were re-pointed at `executeNested` rather than deleted, so a nested route still provably cannot drain the outer owner's credit or capture their in-flight synced tokens.

</details>

### DIFF
--- a/contracts/base/Dispatcher.sol
+++ b/contracts/base/Dispatcher.sol
@@ -7,6 +7,7 @@ import {V4SwapRouter} from '../modules/uniswap/v4/V4SwapRouter.sol';
 import {BytesLib} from '../modules/uniswap/v3/BytesLib.sol';
 import {Payments} from '../modules/Payments.sol';
 import {Constants} from '../libraries/Constants.sol';
+import {NestedUnlock} from '../libraries/NestedUnlock.sol';
 import {PaymentsImmutables} from '../modules/PaymentsImmutables.sol';
 import {V3ToV4Migrator} from '../modules/V3ToV4Migrator.sol';
 import {Commands} from '../libraries/Commands.sol';
@@ -37,11 +38,28 @@ abstract contract Dispatcher is
 
     error InvalidCommandType(uint256 commandType);
     error BalanceTooLow();
+    error NotSelf();
+    /// @notice Thrown when a route reaches V4_SWAP inside a foreign PoolManager unlock without opting in
+    error NestedExecutionNotPermitted();
 
     /// @notice Executes encoded commands along with provided inputs.
     /// @param commands A set of concatenated commands, each 1 byte in length
     /// @param inputs An array of byte strings containing abi encoded inputs for each command
     function execute(bytes calldata commands, bytes[] calldata inputs) external payable virtual;
+
+    /// @notice Runs the v4 actions of a V4_SWAP command inside a PoolManager unlock opened elsewhere
+    /// @param inputs The V4_SWAP command input, re-encoded by the dispatcher
+    /// @dev Only callable by this contract. It exists so that `inputs` arrives as a freshly abi encoded
+    /// calldata copy rather than as a slice of the original transaction calldata. The v4 parameter decoders
+    /// follow offsets without bounding them against the enclosing input, so a slice would let them read
+    /// bytes past `inputs.length` -- bytes that executeSigned never commits to. Encoding the argument makes
+    /// calldata end where the input ends, which is the same guarantee poolManager.unlock() gives the
+    /// ordinary path when it re-encodes for unlockCallback.
+    function executeV4SwapWithinUnlock(bytes calldata inputs) external {
+        if (msg.sender != address(this)) revert NotSelf();
+        (bytes calldata actions, bytes[] calldata params) = inputs.decodeActionsRouterParams();
+        _executeActionsWithoutUnlock(actions, params);
+    }
 
     /// @notice Public view function to be used instead of msg.sender, as the contract performs self-reentrancy and at
     /// times msg.sender == address(this). Instead msgSender() returns the initiator of the lock
@@ -299,8 +317,10 @@ abstract contract Dispatcher is
                     // contract's unlock callback. Opening a new lock via _executeActions would revert with
                     // AlreadyUnlocked, so instead run the v4 actions within the existing lock.
                     if (poolManager.isUnlocked()) {
-                        (bytes calldata actions, bytes[] calldata params) = inputs.decodeActionsRouterParams();
-                        _executeActionsWithoutUnlock(actions, params);
+                        // Nesting shares the router's delta account with every other call in this
+                        // unlock, so it runs only when the caller explicitly opted in via executeNested.
+                        if (!NestedUnlock.isPermitted()) revert NestedExecutionNotPermitted();
+                        this.executeV4SwapWithinUnlock(inputs);
                     } else {
                         // pass the calldata provided to V4SwapRouter._executeActions (defined in BaseActionsRouter)
                         _executeActions(inputs);
--- a/contracts/UniversalRouter.sol
+++ b/contracts/UniversalRouter.sol
@@ -9,6 +9,7 @@ import {PaymentsImmutables, PaymentsParameters} from './modules/PaymentsImmutabl
 import {UniswapImmutables, UniswapParameters} from './modules/uniswap/UniswapImmutables.sol';
 import {V4SwapRouter} from './modules/uniswap/v4/V4SwapRouter.sol';
 import {Commands} from './libraries/Commands.sol';
+import {NestedUnlock} from './libraries/NestedUnlock.sol';
 import {IUniversalRouter} from './interfaces/IUniversalRouter.sol';
 import {MigratorImmutables, MigratorParameters} from './modules/MigratorImmutables.sol';
 import {EIP712} from '@openzeppelin/contracts/utils/cryptography/EIP712.sol';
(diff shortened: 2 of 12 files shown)


contracts/base/Dispatcher.sol
```diff
@@ -299,6 +302,13 @@ abstract contract Dispatcher is
                     // contract's unlock callback. Opening a new lock via _executeActions would revert with
                     // AlreadyUnlocked, so instead run the v4 actions within the existing lock.
                     if (poolManager.isUnlocked()) {
+                        // Nesting shares the router's delta account with every other call in this
+                        // unlock, so it runs only when the caller explicitly opted in via executeNested.
+                        if (!NestedUnlock.isPermitted()) revert NestedExecutionNotPermitted();
+                        // The v4 swap parameter decoders (v4-periphery) bound every struct and member offset
+                        // against the input length via abi.decode, so decoding directly from this calldata
+                        // slice cannot read past inputs[i].length. Coverage: V4NestedMemberSmuggle.t.sol and
+                        // V4NestedUnlockCalldataSmuggling.t.sol.
```
zhongeric: Rm this AI generated comment

PR #503, opened 2026-08-25
docs: record the V2 per-hop bound donation limitation as a known issue

dianakocsis:
Documents audit finding **L-01 (Low)** rather than fixing it. Comments and tests only — no behaviour change, bytecode unchanged.

## The issue

`_v2Swap` derives each hop's input from `balanceOf(pair) - reserveInput`, which counts tokens any third party can transfer to the pair. Because a larger apparent trade earns a worse average rate on a constant-product curve, a donation drives the measured price below `minHopPriceX36` and reverts a bounded route. The pair's permissionless `skim()` returns the donation afterwards, so censorship costs the attacker only gas and transaction ordering.

The perverse part: the donation *improves* the victim's output. They receive more tokens for the same input, and the route reverts anyway.

No funds are at risk, and the bound never lets bad execution through — the failure mode is a spurious revert. Hence Low.

## Why the fix is deferred

- The correct primitive needs routing to emit different per-hop values, so it isn't contained to this repo.
- The encoding migration needs a decision first: the field's *meaning* would change without its *type* changing, so a stale caller would be silently misread rather than rejected. That likely wants new command IDs.
- We're close to the bytecode limit and want to scope this deliberately.

Fixing in the next version.

## What's here

Two PoCs that pass against current `main`, as executable documentation:

**`V2PerHopDonationPoC.t.sol`** — single hop. Shows the donation strictly improves the victim's output when unbounded, censors the identical swap when bounded, and is fully recovered via `skim()`. Also shows attack cost scales with *pool reserves*, not the victim's trade size.

**`V2PerHopDonationMultihop.t.sol`** — two hops, with **hop 0's bound set to zero** and only hop 1 bounded. The route is still censored, at hop 1, whose own pair was never touched.

That second one is the important one: it rules out input attribution as a fix. Hop 0 *swaps* the donation, so the proceeds physically arrive at pair 1 — hop 1 then receives genuinely more input and honestly measures a worse rate. There is nothing foreign left to attribute away. Both the audit's recommended patch and the "hop i>0 denominator = previous hop's measured output" shape fail here.

Measured on a 10,000/10,000 pool with a 50 bps tolerance:

| | |
|---|---|
| victim's trade | 10 ether |
| donation to censor it | 50.96 (**50 bps of pair0's reserves**) |
| B reaching pair1, honest | 9.96 |
| B reaching pair1, attacked | 60.41 |
| donation recovered via `skim` | 100% |

## Related V3 exposure

Also noted in `V3SwapRouter.sol`: when `amountIn == CONTRACT_BALANCE` the router seeds `amountIn` from its own balance, so a donation *to the router* enlarges the trade the same way, and `SWEEP` takes an arbitrary recipient so it's recoverable. This is not in the audit report.

V3's per-hop denominator is itself sound — it divides by the pool's callback `amountToPay`, not a balance. The exposure is only that one mode, and a guard rejecting a nonzero bound under `CONTRACT_BALANCE` would close it. V4 is unaffected; its denominators are all flash-accounting deltas.

## Prototype for next version

A working reference implementation is parked on `fix/v2-hop-bound-from-reserves` (commit `47e3c4c`, **not for merge**). It replaces the ratio with a reserve-quoted bound: each entry packs `(referenceAmountIn, minAmountOut)` and the hop is checked with `getAmountOut(referenceAmountIn, reserveIn, reserveOut)` before the swap. Reading only reserves and calldata makes it immune to donations, keeps each hop independent of upstream drift, and works under `ALREADY_PAID` / `CONTRACT_BALANCE`. It also drops **282 bytes** of bytecode (23,705 → 23,423). 8 tests, all passing.

Its known gaps are recorded in that commit message.

## Verification

`forge test` — 88 passed, 0 failed (excluding the pre-existing Celo fork-test failures on `main`, which fail identically without this change).

<details>

Documents audit finding **L-01 (Low)** rather than fixing it. Comments and tests only — no behaviour change, bytecode unchanged.
## The issue
`_v2Swap` derives each hop's input from `balanceOf(pair) - reserveInput`, which counts tokens any third party can transfer to the pair. Because a larger apparent trade earns a worse average rate on a constant-product curve, a donation drives the measured price below `minHopPriceX36` and reverts a bounded route. The pair's permissionless `skim()` returns the donation afterwards, so censorship costs the attacker only gas and transaction ordering.
No funds are at risk, and the bound never lets bad execution through — the failure mode is a spurious revert. Hence Low.
## Changes
- **`V2SwapRouter.sol`**: Add inline `KNOWN ISSUE` comments at the `amountInput` derivation and NatSpec on `v2SwapExactInput` / `v2SwapExactOutput` documenting the donation-censorship vector and why input attribution does not close multihop routes
- **`V3SwapRouter.sol`**: Add inline `KNOWN ISSUE` comment at the `CONTRACT_BALANCE` branch noting the related V3 exposure (donation to the router enlarges the trade; `SWEEP` recovers it)
- **`V2PerHopDonationPoC.t.sol`**: Single-hop PoC — shows the donation strictly improves the victim's output when unbounded, censors the identical swap when bounded, is fully recovered via `skim()`, and that attack cost scales with pool reserves not trade size
- **`V2PerHopDonationMultihop.t.sol`**: Two-hop PoC with hop 0's bound disabled and only hop 1 bounded — route is censored at hop 1 whose own pair was never touched, ruling out input attribution as a fix
- **`MockV2Pair.sol`**: Minimal faithful port of UniswapV2Pair swap/skim/sync/getReserves with a CREATE2 factory matching the `pairFor` derivation scheme
## Why the fix is deferred
- The correct primitive needs routing to emit different per-hop values, so it isn't contained to this repo
- The encoding migration needs a decision first: the field's meaning would change without its type changing, so a stale caller would be silently misread rather than rejected — likely wants new command IDs
- Close to the bytecode limit and want to scope deliberately
## Notes
- A working reference implementation is parked on `fix/v2-hop-bound-from-reserves` (commit `47e3c4c`, not for merge) — replaces the ratio with a reserve-quoted bound, drops 282 bytes of bytecode
- `forge test` — 88 passed, 0 failed (excluding pre-existing Celo fork-test failures on `main`)

</details>

### DIFF
--- a/contracts/modules/uniswap/v2/V2SwapRouter.sol
+++ b/contracts/modules/uniswap/v2/V2SwapRouter.sol
@@ -30,6 +30,14 @@ abstract contract V2SwapRouter is UniswapImmutables, Permit2Payments {
                 (uint256 reserve0, uint256 reserve1,) = IUniswapV2Pair(pair).getReserves();
                 (uint256 reserveInput, uint256 reserveOutput) =
                     input == token0 ? (reserve0, reserve1) : (reserve1, reserve0);
+                // KNOWN ISSUE (audit L-01, fix deferred): amountInput is the pair's whole balance
+                // excess, so it counts tokens any third party transferred in. A larger apparent trade
+                // earns a worse average rate, so a donation drives the price below minHopPriceX36 and
+                // reverts a bounded route; the pair's permissionless skim() then returns the donation.
+                // No funds are at risk, the effect is censorship of routes that set a nonzero bound.
+                // Attributing the caller's own input does not close it on multihop routes: hop 0 swaps
+                // the donation, so hop 1 receives genuinely more input and honestly measures a worse
+                // rate. See test/foundry-tests/V2PerHopDonationMultihop.t.sol.
                 uint256 amountInput = ERC20(input).balanceOf(pair) - reserveInput;
                 uint256 amountOutput = UniswapV2Library.getAmountOut(amountInput, reserveInput, reserveOutput);
                 (uint256 amount0Out, uint256 amount1Out) =
@@ -64,6 +72,9 @@ abstract contract V2SwapRouter is UniswapImmutables, Permit2Payments {
     /// @param path The path of the trade as an array of token addresses
     /// @param payer The address that will be paying the input
     /// @param minHopPriceX36 Per-hop minimum price array in 1e36 precision (empty to disable)
+    /// @dev KNOWN ISSUE (audit L-01, fix deferred): a nonzero bound can be tripped by anyone
+    /// transferring tokens to a pair in the path, censoring the route at no net cost to them. See
+    /// the note in _v2Swap.
     function v2SwapExactInput(
         address recipient,
         uint256 amountIn,
@@ -101,6 +112,9 @@ abstract contract V2SwapRouter is UniswapImmutables, Permit2Payments {
     /// @param path The path of the trade as an array of token addresses
     /// @param payer The address that will be paying the input
     /// @param minHopPriceX36 Per-hop minimum price array in 1e36 precision (empty to disable)
+    /// @dev KNOWN ISSUE (audit L-01, fix deferred): a nonzero bound can be tripped by anyone
+    /// transferring tokens to a pair in the path, censoring the route at no net cost to them. See
+    /// the note in _v2Swap.
     function v2SwapExactOutput(
         address recipient,
         uint256 amountOut,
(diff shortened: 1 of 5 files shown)


dianakocsis:
Closing — opened by mistake, the work just needed to live on a branch for now. L-01 is documented and deliberately not fixed in this release; branch `docs/v2-per-hop-bound-known-issue` keeps the notes and PoCs, and `fix/v2-hop-bound-from-reserves` (`47e3c4c`) keeps the prototype for the next version.

PR #510, opened 2026-09-09
refactor: drop the nested-unlock calldata re-encode, redundant under #505

ccashwell:
## What

Removes `executeV4SwapWithinUnlock`, the self-call #502 added so the nested `V4_SWAP` decoded a fresh abi-encoded copy of its input instead of a slice of the transaction calldata. The nested branch now decodes directly from its calldata slice, the same way it did before #502.

## Why it is safe to remove

The self-call existed to cap `calldatasize()` at the input boundary, which mattered only because the old assembly swap-param decoders followed struct and member offsets bounded by `calldatasize()` rather than by `params.length`. #505 bumped v4-periphery to decoders that use `abi.decode`, which bounds the struct offset and every dynamic member against the input length. So the decoders now end the read at the input boundary on their own, on the nested path as much as the ordinary one, and the re-encode is redundant.

## Verification

A three-way build matrix drove the member-offset smuggle (a redirected `path` member, the variant a struct-offset-only fix misses, where the whole `PathKey` including the hooks address escapes):

| Decoders | re-encode | member-offset smuggle |
|---|---|---|
| #502 assembly | removed | **executes** (control: exploit is real) |
| #505 `abi.decode` | removed | **blocked** |
| #505 `abi.decode` | present | **blocked** |

The control routed a swap through a path sourced entirely from bytes past `inputs[0].length` and delivered `996006981039903` out, matching the figure in the original finding. New regression coverage is `V4NestedMemberSmuggle.t.sol`; the existing struct-offset tests in `V4NestedUnlockCalldataSmuggling.t.sol` are unchanged and still pass, and the obsolete self-call test is dropped with the function.

## Tradeoff

The only property lost is defense in depth: if a future v4-periphery action decoder reverts to an unbounded assembly pointer, the re-encode would have kept the nested path safe without anyone re-noticing. That is a forward-looking cost, not a current exposure.

## Bytecode

| | Runtime | Free under 24,576 |
|---|---|---|
| v4 poke PR | 24,390 | 186 |
| This PR | 24,243 | **333** |

Frees the room the v3 protocol fee command needs.

<details>

## What
Removes `executeV4SwapWithinUnlock`, the self-call #502 added so the nested `V4_SWAP` decoded a fresh abi-encoded copy of its input instead of a slice of the transaction calldata. The nested branch now decodes directly from its calldata slice, the same way it did before #502.
## Why it is safe to remove
The self-call existed to cap `calldatasize()` at the input boundary, which mattered only because the old assembly swap-param decoders followed struct and member offsets bounded by `calldatasize()` rather than by `params.length`. #505 bumped v4-periphery to decoders that use `abi.decode`, which bounds the struct offset and every dynamic member against the input length. So the decoders now end the read at the input boundary on their own, on the nested path as much as the ordinary one, and the re-encode is redundant.
## Changes
- **`Dispatcher.sol`**: Remove `executeV4SwapWithinUnlock` and its `NotSelf` error; inline the `decodeActionsRouterParams` + `_executeActionsWithoutUnlock` calls directly in the nested `V4_SWAP` branch
- **`V4NestedMemberSmuggle.t.sol`**: New test covering member-offset smuggling (redirected `path` member, the variant a struct-offset-only fix misses); confirms the `abi.decode` decoders block it
- **`V4NestedUnlockCalldataSmuggling.t.sol`**: Update NatSpec to reflect the new safety model (decoder-bounded vs re-encode); drop the `test_executeV4SwapWithinUnlock_revertsForExternalCallers` test alongside the removed function
- **`UniversalRouterTest.json`**: Bytecode size 24390 → 24243
## Verification
A three-way build matrix drove the member-offset smuggle (a redirected `path` member, the variant a struct-offset-only fix misses, where the whole `PathKey` including the hooks address escapes):
| Decoders | re-encode | member-offset smuggle |
|---|---|---|
| #502 assembly | removed | **executes** (control: exploit is real) |
| #505 `abi.decode` | removed | **blocked** |
| #505 `abi.decode` | present | **blocked** |
## Bytecode
| | Runtime | Free under 24,576 |
|---|---|---|
| v4 poke PR | 24,390 | 186 |
| This PR | 24,243 | **333** |
Frees the room the v3 protocol fee command needs.

</details>

### DIFF
--- a/contracts/base/Dispatcher.sol
+++ b/contracts/base/Dispatcher.sol
@@ -42,7 +42,6 @@ abstract contract Dispatcher is
 
     error InvalidCommandType(uint256 commandType);
     error BalanceTooLow();
-    error NotSelf();
     /// @notice Thrown when a route reaches V4_SWAP inside a foreign PoolManager unlock without opting in
     error NestedExecutionNotPermitted();
 
@@ -50,20 +49,6 @@ abstract contract Dispatcher is
     /// @param commands A set of concatenated commands, each 1 byte in length
     /// @param inputs An array of byte strings containing abi encoded inputs for each command
     function execute(bytes calldata commands, bytes[] calldata inputs) external payable virtual;
-
-    /// @notice Runs the v4 actions of a V4_SWAP command inside a PoolManager unlock opened elsewhere
-    /// @param inputs The V4_SWAP command input, re-encoded by the dispatcher
-    /// @dev Only callable by this contract. It exists so that `inputs` arrives as a freshly abi encoded
-    /// calldata copy rather than as a slice of the original transaction calldata. The v4 parameter decoders
-    /// follow offsets without bounding them against the enclosing input, so a slice would let them read
-    /// bytes past `inputs.length` -- bytes that executeSigned never commits to. Encoding the argument makes
-    /// calldata end where the input ends, which is the same guarantee poolManager.unlock() gives the
-    /// ordinary path when it re-encodes for unlockCallback.
-    function executeV4SwapWithinUnlock(bytes calldata inputs) external {
-        if (msg.sender != address(this)) revert NotSelf();
-        (bytes calldata actions, bytes[] calldata params) = inputs.decodeActionsRouterParams();
-        _executeActionsWithoutUnlock(actions, params);
-    }
 
     /// @notice Public view function to be used instead of msg.sender, as the contract performs self-reentrancy and at
     /// times msg.sender == address(this). Instead msgSender() returns the initiator of the lock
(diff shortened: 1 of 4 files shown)


dianakocsis:
Superseded by #513.

#513 is this change rebased onto #502 directly, so it no longer depends on #507, #508 or #509. Verified equivalent before closing:

- `V4NestedMemberSmuggle.t.sol` and `V4NestedUnlockCalldataSmuggling.t.sol` are byte-for-byte identical (sha-matched)
- The `Dispatcher.sol` change is the identical semantic diff — drop `NotSelf`, drop `executeV4SwapWithinUnlock` + NatSpec, replace the self-call with the direct decode
- Same four files touched, no extras
- Both remove exactly 147 bytes of runtime (here 24,390 -> 24,243; on #502 24,500 -> 24,353)

#513 also carries what this PR was missing: the Hardhat gas snapshots, which were never regenerated here and are why `Unit Tests` sat at 31 failures. On #513 all 31 entries fall by exactly 22 gas, or 44 where a sub-plan enters the router twice — one fewer comparison in the external selector dispatch table. #513 is green on all checks.

The safety argument is unchanged and was re-verified directly on #502's decoders (`0cc6e164`, abi.decode) rather than inferred from the three-way matrix: the member-offset smuggle is blocked. Note that #502's pre-existing configuration — assembly decoders with the re-encode present — was the one row this PR's matrix never tested, and #513 fills it in.

**Not deleting the branch:** #511 is still based on `refactor/drop-nested-reencode`. It needs rebasing onto #502 before this branch can go.

PR #470, opened 2026-02-21
feat: Universal Router v2.1.1

gretzke:
## Summary
- **PAY_PORTION_FULL_PRECISION** (`0x07`): New payment command with 1e18 precision (vs 10,000 bips), enabling exact fractional payments like 1/3
- **Per-hop slippage for V2/V3**: Adds `uint256[] maxHopSlippage` parameter to all 4 V2/V3 swap commands (`V3_SWAP_EXACT_IN`, `V3_SWAP_EXACT_OUT`, `V2_SWAP_EXACT_IN`, `V2_SWAP_EXACT_OUT`), matching V4's convention where `price = amountOut * 1e36 / amountIn`
- **v4-periphery submodule bump**: Updates submodule to include `maxHopSlippage` in V4 single-swap structs

## Test Plan
- [x] 101 Foundry tests pass (including 13 new V3 mock pool tests for per-hop slippage)
- [x] 12 Hardhat integration tests pass (including new tests for PAY_PORTION_FULL_PRECISION)
- [x] V3 exact-output multi-hop slippage enforcement verified (hop 0 bug fixed)
- [x] V4 single-swap planner ABI updated for maxHopSlippage field
- [x] Path length guards prevent underflow on malformed inputs

---
## Summary
Universal Router v2.1.1 introduces high-precision payment commands and per-hop slippage protection for V2/V3 swaps, aligning with V4's price convention.
### New Commands
- **PAY_PORTION_FULL_PRECISION** (`0x07`): Payment command with 1e18 precision (vs 10,000 bips), enabling exact fractional payments like 1/3
### Per-hop Slippage Protection
- Add `uint256[] minHopPriceX36` parameter to all V2/V3 swap commands:
  - `V3_SWAP_EXACT_IN`, `V3_SWAP_EXACT_OUT`
  - `V2_SWAP_EXACT_IN`, `V2_SWAP_EXACT_OUT`
- Price convention: `amountOut * 1e36 / amountIn`
- V2 swaps measure actual balance changes for accurate FOT token handling
- Path length guards prevent underflow on malformed inputs
### New Errors
- `V2TooLittleReceivedPerHop`, `V2InvalidHopPriceLength` in V2SwapRouter
- `V3TooLittleReceivedPerHop`, `V3TooMuchRequestedPerHop`, `V3HopPriceAndPathLengthMismatch` in V3SwapRouter
- `InvalidPortion` in Payments module
### Library Updates
[...]

PR #515, opened 2026-09-16
Deploy 2: pin v4-periphery to #564 for the 2.1.x redeploy

dianakocsis:
## Base branch

This targets **`release/2.1.1-base`**, a frozen pointer at `999d561` (tag `2.1.1`), *not* `main`. Tags can't be PR bases, so that branch exists purely to give GitHub a selectable anchor — it never moves. Basing on `main` would render an 83-file, −1,851-line diff that looks like tearing out 2.2.0.

This release line is deliberately **never merged into `main`**. `main` stays on 2.2.x. The branch is tagged and deployed from directly.

## What changed

Four commits, 10 files, **zero production contract code**:

| file | change |
|---|---|
| `lib/v4-periphery` | `3231810e` → `545a5d2a` (#564, revert on exact-output partial fills) |
| `foundry.toml` | `UniversalRouter` compiles at `optimizer_runs = 3000` |
| `hardhat.config.ts` | `runs: 1` → `3000` to match |
| `test/foundry-tests/SwapProxy.t.sol` | fixture fix (7 encodings) |
| `snapshots/UniversalRouterTest.json` | 24,546 → **24,380** |
| `test/integration-tests/gas-tests/__snapshots__/*` | 65 values regenerated |

## Why the optimizer change is required

#564 adds ~158 bytes to `V4Router`, which compiles into `UniversalRouter`. Tag `2.1.1` shipped at 24,546 bytes with **30 bytes of headroom**, so the pin bump alone lands at **24,704 — 128 bytes over the EIP-170 limit and undeployable**.

`3000` is the highest value that fits:

| runs | bytes | margin |
|---|---:|---:|
| 1 | 20,900 | 3,676 |
| 1000 | 23,270 | 1,306 |
| 2000 | 23,994 | 582 |
| **3000** | **24,380** | **196** |
| 3500 | 24,587 | −11 |
| 4444 | 24,704 | −128 |

## Gas impact

At `3000` the cost versus the `4444` that the live 2.1.1 router (`0x4C82D1fB…`) was deployed with is **effectively nil** — measured across 102 forge tests, built both ways:

| suite | vs 4444 |
|---|---:|
| V2 suites, `RouteSignerTest`, `ChainedActionsFork` | **+0.00%** (byte-identical) |
| `SwapProxyTest` | +0.02% |
| `UniversalRouterTest` | +0.01% |
| `UniswapV3Test` | +0.14% |
| **overall median** | **+0.00%** |

Worst individual case is +484 gas (+0.16%) on V3 multihop exact-output — about $0.00005 at current fees. For comparison, `runs=1` would have cost +1.00% median and +1.49% on V3.

`hardhat.config.ts` also moves to 3000. It had been pinned at `1` since at least `8c79e61` while forge built at `4444`, so the committed gas snapshots described a build that was never deployed. Both toolchains now agree and the regenerated snapshots reflect the artifact that actually ships.

## Why the SwapProxy fixture fix is required

The four `SwapProxy` swap tests have failed with `SliceOutOfBounds()` since #469 merged — CI recorded a **failure** on `999d561` and it was merged red. Cause: #470 added a trailing per-hop price bound array to the V2/V3 command encoding, and #469's fixtures were written without it, so the decoder overruns. The fix is the fixture change from #476, which repaired the same tests on the 2.2.0 line. **This branch therefore has a greener suite than tag `2.1.1` itself.**

## Verification

- `forge build --sizes` — **24,380 bytes, 196 free**
- `forge test --isolate` — **105 passed, 0 failed** (was 101/4 at the tag)
- `hardhat test` — **184 passing, 2 pending**, snapshots compared not updated
- `prettier --check` — clean
- Deploy dry run against mainnet succeeds; all 10 constructor params verified against the live 2.1.1 contract onchain

## Deployment notes

`lib/v4-periphery` is pinned at `545a5d2a`, which contains #564 and #574/#577 but **not** #584, #585, #586 or #587.

Tag to be applied to the deployed commit; the release is deployed from this branch rather than `main`, so the tag and a `deploy-addresses` entry recording address + commit SHA are the only durable provenance.


ericneil-sanc (approved):

dianakocsis:
Plan: tag: 2.1.2 on 802fe4c before deploying.
Then close the PR. Will not merge

dianakocsis:
Closing rather than merging. this PR's deliverable was the tag and the deployments, not the diff.
Tag `2.1.2` was cut at 802fe4c (this branch's head), and that commit is what all 24 chains were built from. Release: https://github.com/Uniswap/universal-router/releases/tag/2.1.2

The address records and deploy-script fixes landed separately in [#516](https://github.com/Uniswap/universal-router/pull/516).

